PR #24, opened 2015-07-27
[WIP] remove gil

jmschrei:
This branch currently makes all HMM methods at least 2 times faster. Distributions range from being 20% faster (uniform distribution) to on average 2x faster (Gaussian, Exponential...), to 30x faster (kernel densities). The forward/backward algorithms are about twice as fast as well, and the gil is released for many functions in an attempt to add multithreaded training.

However, I've hit a road block as the gil must be reacquired to interact with distribution objects since not all of them (in particular, DiscreteDistributions) can have the gil removed nicely.

### DIFF
--- a/pomegranate/base.pxd
+++ b/pomegranate/base.pxd
@@ -13,3 +13,4 @@ cdef class Model( object ):
 	cdef public str name
 	cdef public list states, edges
 	cdef public object graph
+	cdef int n_edges, n_states
--- a/pomegranate/base.pxd
+++ b/pomegranate/base.pxd
@@ -12,7 +12,4 @@ cdef class State( object ):
 cdef class Model( object ):
 	cdef public str name
 	cdef public list states, edges
-	cdef object graph
-	cdef int [:] in_edge_count, in_transitions, out_edge_count, out_transitions
-	cdef double [:] in_transition_log_probabilities
-	cdef double [:] out_transition_log_probabilities
+	cdef public object graph
--- a/pomegranate/base.pyx
+++ b/pomegranate/base.pyx
@@ -158,6 +158,8 @@ cdef class Model( object ):
 		self.name = name or str( id(self) )
 		self.states = []
 		self.edges = []
+		self.n_edges = 0
+		self.n_states = 0
 	
 	def __str__(self):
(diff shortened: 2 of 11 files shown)


aaiyer:
I would create a Cython only implementation of all the classes (that doesn't hold the gil) and have wrapper classes around the "cdef classes" .. Then, whenyou take an input that is say a distribution, you grab the underlying Cython class in the wrapper only. Whenever you need to return a Cython class, just wrap it and return.

I hope my explanation makes sense.

jmschrei:
Hey, thanks for the help. I'm not sure that I understand, though. Can you provide an example?

aaiyer:
Sure.

Let says I have a Cython class:

```
cdef class AImpl:
    def __init__(self) nogil:
         pass
    def some_func(self) nogil:
         pass
```

I wrap it using a Python class
```
cpdef class A:
    cdef AImpl __impl
    def __init__(self, impl):
        self.__impl = impl
    def some_func(self):
        self.__impl.some_func()
```
Now, I want to pass it to another class,

```
cdef BImpl:
     cdef AImpl a
     def __init__(self, A a):
          self.a = a.__impl
     
    def do_something_on_a(self) nogil:
        self.a.some_func()

    def give_me_a(self):
        return A(self.a) # Wrap it when you return it
```

```
cpdef class B:
    cdef BImpl __impl
    def __init__(self, impl):
        self.__impl = impl
    def do_something_on_a(self) nogil:
        self.__impl.do_something_on_a()
    def give_me_a(self):
        return self.__impl.give_me_a()
````

Hope that explains it :)

jmschrei:
I'm not entirely sure where this would help?

aaiyer:
Avoid the GIL totally when training as you will only be dealing with Cython
classes. You commented that you were having trouble.

aaiyer:
Particularly you will be holding reference to Cython classes when you require distribution objects. As you will be interacting only with Cython classes, you can avoid the GIL. The only thing you need to do is in the Distribution objects, is replace the list and dict objects with cython.vector or a c++ map

jmschrei:
The other two PRs have covered this PR. Closing.

PR #25, opened 2015-08-05
HMM: Sync transition probabilities back into graph's edges after training. Better visualization.

aaiyer:
* Important bug fix -- Sync transition probabilities back into graph's edges after training. If not we lose it when serializing the model.
 * Nicer visualization
 * State names must be unique

### DIFF
--- a/pomegranate/hmm.pyx
+++ b/pomegranate/hmm.pyx
@@ -2,11 +2,14 @@
 # Contact: Jacob Schreiber ( jmschreiber91@gmail.com )
 #          Adam Novak ( anovak1@ucsc.edu )
 
+from __future__ import print_function
+
 cimport cython
 from cython.view cimport array as cvarray
 from libc.math cimport log as clog, sqrt as csqrt, exp as cexp
 import math, random, itertools as it, sys, bisect, json
 import networkx
+import tempfile
 
 if sys.version_info[0] > 2:
 	# Set up for Python 3
@@ -29,6 +32,8 @@ cimport base
 from base cimport *
 
 from matplotlib import pyplot
+import matplotlib.image
+import pygraphviz
 
 # Define some useful constants
 DEF NEGINF = float("-inf")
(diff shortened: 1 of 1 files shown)


jmschrei:
Awesome, I'll take a look at this in more depth a little later. Can you show an example of what the visualizations look like now on some simple models?

aaiyer:
Here is a visualization of the model in test_hmm_training.py


figure_1

jmschrei:
That's so awesome. I'm a bit hesitant to add pygraphviz as a ~hard~ requirement. Can you add a small patch to check and see if the person has it, and if so, run as you've patched, and otherwise (when they try to draw) draw the graph as normal and raise a Warning (not an error) suggesting that they install them for more meaningful graphs? I think that will fix the Travis issue as well.

aaiyer:
Done :)

aaiyer:
Fixed the transition probabilities labels, graph looks even better:

figure_10

jmschrei:
I like how frozen states are blue. =)

Can you elaborate on the use of in_probability_transition_edge_index? Is this necessary, versus the current (albiet more complex) indexing for sparse matrices?

aaiyer:
in_probability_transition_edge_index is just used to sync the transition probabilities from in_transition_log_probabilities back to the edges in the graph. i could use the more complex way from bake(), but that would be a repetition of code. factoring out that block of code from bake() would make it even harder to read IMHO.

jmschrei:
I've reviewed your PR more thoroughly! It looks very helpful and so I'd like to merge it ASAP, but has a few issues.

* You correctly identify that I fail to update the hmm.graph model after training. This needs to be fixed
* The way you do it, you only update the in edge probabilities, not the out edge ones as well
* Your edge_index dictionary works when the GIL is not released, but in my other PR where the GIL is being released to support multi-threaded training, this will not be viable anymore
* There is a way to update the model using existing structures that do not require the GIL

My suggestion is that for now, you open another PR where you only add in the better plotting option. After we merge that, I can tag in and patch the hmm.graph attribute after training.

As for the readability of the sparse data structures, it's a bit unintuitive to those who don't already know about them, but it's a standard (and extremely efficient) way to represent a sparse matrix (through the use of three vectors), called CSR. 

My laptop has unfortunately died, and so my development of personal projects has slowed down recently. =/

aaiyer:
Will get the person who gave me the patch (Joe McIntyre) to open a PR for the visualization.

PR #27, opened 2015-08-06
Small mistake

one541:
Hi Jacob,

I accidentally introduced a simple dumb mistakes in my previous PR.

-Joe


jmschrei:
Whoops, I should've caught that. Thanks for the patch.

PR #29, opened 2015-09-12
[WIP] BayesianNetwork imputation

jmschrei:
This pull requests add an imputation method to BayesianNetworks, which allows them to take in a matrix of data with missing values, and impute the most likely value given the structure and emissions of the network.

- [x] Add imputation method
- [x] Add test case
- [x] Add an example of usage


jmschrei:
Merging.

PR #33, opened 2015-09-27
[MRG] ENH imports reduced

jmschrei:
I've significantly cleaned up the tops of the files, removing extraneous imports which had built up because I simply copied the headers when starting new files. No new functionality is proposed. I've also added relative imports instead of absolute imports for imports from within this package, to make the logic easier to understand.


jmschrei:
Checks passing. Merging.

PR #34, opened 2015-09-27
[WIP] unit test failureson python 3

jmschrei:


jmschrei:
So it looks like the error isn't 3.5, the error is that there is a random failing in the unit tests related to Conditional Probability Tables. I've made some progress narrowing down the issue, but don't know what the cause is. Lame.

jmschrei:
I rewrote Conditional Probability Tables to be safer, simpler to understand the code (and slightly faster for log probability calculations), and so now the first Bayesian network tests don't fail. However, the forward backward tests do fail because of an error with DiscreteDistributions. This only happens in Python 3, not Python2.7, however. I'm guessing one of the assistance functions isn't safe.

jmschrei:
I did not figure out explicitly what the error was, other than that it came from `JointProbabilityTable.marginal`. After refactoring it to have the new API that CPTs have, the code works (and is cleaner to understand.) Good riddance, random bugs on deterministic tests!

PR #37, opened 2015-10-12
ENH removed gil requiring time

jmschrei:
This passes the distributions as a void array so that we don't need to acquire the gil inside Baum-Welch to unpack them. It also removes acquiring the gil to modify the expected transitions matrix. This performs slightly better, increasing speed from ~x1.8 to ~x2.4 faster with 4 cores. There is more work to be done, but this is another good step in the right direction.


jmschrei:
merged

PR #38, opened 2015-10-13
[MRG] hmm sparse expected transition matrix

jmschrei:
More updates to make multithreading more efficient. Now the internal expected transitions matrix has been reduced down to a vector for sparse format. This reduces the amount of time spent with the gil acquired at the end of each round. Also, improvements have been made to `model.log_probability`, which seems to be the primary bottleneck.


jmschrei:
We get up to a 3x improvement with 4 cores. Still not the best, but I'm trying to track down where the bottleneck could be. The majority of the code has the gil released, so it has to either be in the few lines of non-thread-released code, or in using the thread pool, since I am sharing it between two tasks.

PR #61, opened 2016-01-09
ENH ipython notebooks

neonrights:
created ipython notebooks for each example and deleted python code for each example.


jmschrei:
Looks mostly good. I'll show you how to squash the commits into one on Tuesday.

PR #103, opened 2016-03-24
API Reference on Website

jmschrei:
Use recommonmark to add API references to the website.


jmschrei:
Looks like it's working. API References have been added for all models on the website.

jmschrei:
The navigation bar is set up as well. Looks like the website is a go.

PR #118, opened 2016-04-11
[ENH] EM speed enhance

jmschrei:
Utilizing caching of the log probability during EM calculations, this PR will improve the speed of both GMMs and HMMs by around 33% during training.


jmschrei:
Unfortunately it also causes segfaults in the Viterbi training, so...

jmschrei:
Finally passing. This was the worst.

PR #119, opened 2016-04-14
Adding distance change report to FactorGraph forward/backward algorithm

kellrott:
Adding some additional logging to the verbose option of the forward_backward algorithm. This includes reporting the iteration and the max distribution 'distance' change.


jmschrei:
Hey, thanks a lot for the contribution! What's the benefit of using euclidean distance for this measurement versus Kullback-Leibler divergence?

jmschrei:
Also, can we make the `if verbose` print statement a single one per line? something like

```
if verbose:
    print "Iteration {}: Euclidean Distance Change {}".format( iteration, distance )
```

Thanks!

PR #124, opened 2016-04-26
Fix GMM example more

danstowell:
pomegranate seems to require 2D data arrays.

Someone please confirm that my edit here matches what the documentation author intended...


jmschrei:
Thanks. You used to be able to pass in 1d arrays, but I decided it wasn't worth it.

PR #127, opened 2016-05-05
Fix memory leak

fcaldas:
Adds 4 frees to the end of _viterbi() function clearing allocated memory


jmschrei:
There are a large number of aesthetic changes here as well removing white space. Can you get rid of those and just add in the free lines?

fcaldas:
Sure, I forgot to disable a plugin that removes trailing whitespaces, sorry

fcaldas:
Also, there is a return in the middle of the call for _viterbi() function, so I added a few calls for free() just before.

jmschrei:
Great, thanks!

PR #130, opened 2016-05-10
Added error handling for naive bayes

neonrights:
Added error handling to check for right model and data inputs. Updated unit-test to reflect these changes and to test the error handling.

Added a simple helper method to convert input data to a usable numpy.ndarray.

Added a new public variable, NaiveBayes.d, that holds the dimension of the models.

Currently passes all naive bayes unit tests.

### DIFF
--- a/pomegranate/NaiveBayes.pyx
+++ b/pomegranate/NaiveBayes.pyx
@@ -13,6 +13,7 @@ from libc.math cimport exp as cexp
 from .distributions cimport Distribution
 from .hmm import HiddenMarkovModel
 from .utils cimport pair_lse
+from .utils import _convert
 
 cdef double NEGINF = float("-inf")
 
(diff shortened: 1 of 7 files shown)


pomegranate/NaiveBayes.pyx
```diff
@@ -58,23 +58,40 @@ cdef class NaiveBayes( object ):
...
         self.summaries = None
-        self.initialized = False
-
-        if not callable(models):
-            self.initialized = True
+        self.d = -1
+        
+        if type(models) is list:
+            if not callable( models[0] ):
+                if isinstance( models[0], HiddenMarkovModel ):
```
jmschrei: What is the reason for this? HiddenMarkovModel should have dimensionality too

pomegranate/NaiveBayes.pyx
```diff
@@ -58,23 +58,40 @@ cdef class NaiveBayes( object ):
...
-
+    
     cdef int initialized
     cdef public object models
     cdef void** models_ptr
     cdef numpy.ndarray summaries
     cdef public numpy.ndarray weights
     cdef double* weights_ptr
+    # dimension of inputs, 0 is dimensionless, -1 if uninitialized
+    cdef public int d
```
jmschrei: Since this has a defined meaning in all other classes, it should be consistent, with d meaning the number of dimensions the model considers.

pomegranate/NaiveBayes.pyx
```diff
@@ -58,23 +58,40 @@ cdef class NaiveBayes( object ):
...
+                    self.d = 0
+                else:
+                    self.d = models[0].d
+            else:
+                raise TypeError("must have initialized models in list")
+
+            for model in models:
+                if callable(model):
+                    raise TypeError("must have initialized models in list")
+                elif self.d == 0 and not isinstance( model, HiddenMarkovModel ):
```
jmschrei: This isn't necessarily true either. You should be able to compare a HMM to a Bayes net or GMM if they have the same dimensionality, aka a sequence of length 3 in 1 dimension can be considered the same as a data point of dimension 3, it's just that the HMM imposes structure whereas the others assume independence.

pomegranate/NaiveBayes.pyx
```diff
@@ -320,3 +369,13 @@ cdef class NaiveBayes( object ):
                     y_ptr[i] = j
 
         return y
+
+    cdef numpy.ndarray _convert( self, data ):
```
jmschrei: This probably shouldn't be a cdef function if you aren't using any C level functionality. We might also want to talk about factoring this to be a general utils.pyx function. Good implementation though.

pomegranate/NaiveBayes.pyx
```diff
@@ -138,8 +150,14 @@ cdef class NaiveBayes( object ):
...
-        X = numpy.array(X)
-        y = numpy.array(y)
+        X = _convert(X)
+        y = _convert(y)
+
+        if not isinstance( self.models[0], HiddenMarkovModel ):
+            if X.ndim > 2:
+                raise ValueError("input data has too many dimensions")
+            elif X.ndim == 2 and self.d != X.shape[1]:
+                raise ValueError("input data rows do not match model dimension")
```
jmschrei: For 1d input, should check to make sure that the data matches the model dimension too, e.g. `elif X.ndim == 1 and self.d > 1` or something

PR #131, opened 2016-05-19
ENH added basic json and print methods to naive bayes

neonrights:
created a simple to and from json method as well as a print method for naive bayes.

### DIFF
--- a/pomegranate/NaiveBayes.pyx
+++ b/pomegranate/NaiveBayes.pyx
@@ -13,6 +13,7 @@ from libc.math cimport exp as cexp
 from .distributions cimport Distribution
 from .hmm import HiddenMarkovModel
 from .utils cimport pair_lse
+import json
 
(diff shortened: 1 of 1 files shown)


pomegranate/NaiveBayes.pyx
```diff
@@ -320,3 +321,21 @@ cdef class NaiveBayes( object ):
...
+        nb = {
+            'class' : 'NaiveBayes',
+            'models' : [ json.loads( model.to_json() ) for model in self.models ],
+            'weights' : self.weights.tolist()
+        }
+
+        return json.dumps( nb, separators=separators, indent=indent )
+
+    def from_json( cls, s ):
+        d = json.loads( s )
```
jmschrei: Can you add in the ability to read from a file if a filename is given as well? Otherwise looks good to go

PR #146, opened 2016-06-21
fix memory leak in hmm.forward_backward

wiehan-a:


jmschrei:
Great catch. If all the unit tests pass then I'll merge.

jmschrei:
LTGM, thanks!

PR #147, opened 2016-06-22
fix memory leak in hmm.bake() and discretedistribution.encode()

wiehan-a:
Hi,

This fixes a memory leak I encounter when an HMM object is baked more than once, like in the code below, where `model.copy()` also performs a `bake()` behind the scenes.

```python
models = []
for _ in xrange(5):
    model = HiddenMarkovModel()
    state = State(DiscreteDistribution({"a" : 0.25, "b": 0.25, "c" : 0.25, "d": 0.25}), name="a")
    model.add_transition(model.start, state, 1.0)
    model.add_transition(state, state, 1.0)
    model.add_transition(state, model.end, 1.0)
    model.bake()
    models.append(model)

def concat_models(models):
    new_model = models[0].copy()
    for idx in xrange(1, len(models)):
        model_copy = models[idx].copy()
        new_model.concatenate(model_copy, str(idx))
    new_model.bake()
    return new_model

while True:
    model = concat_models(models)
```

### DIFF
--- a/pomegranate/distributions.pyx
+++ b/pomegranate/distributions.pyx
@@ -1396,6 +1396,10 @@ cdef class DiscreteDistribution( Distribution ):
 
 		n = len(encoded_keys)
 		self.encoded_keys = encoded_keys
+
+		free(self.encoded_counts)
+		free(self.encoded_log_probability)
+
 		self.encoded_counts = <double*> calloc( n, sizeof(double) )
 		self.encoded_log_probability = <double*> calloc( n, sizeof(double) )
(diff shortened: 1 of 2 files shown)


jmschrei:
Great, thanks for the contribution. I am unsure if I want the API to look like that, but until I decide this fixes the issue well enough.

wiehan-a:
Hi,

Fair enough. At the very least `free_bake_buffers` should be mangled and hidden from the API (`__free_bake_buffers`). 

Regards

Issue #153, opened 2016-07-27
Support pickling on hmm


pixelou:
Hello,

I would like to know if you could improve the support for serialization toward something more standard than the `clone()` method for hmms.
Correct me if I am wrong, but the clone() is specific to scikit learn, which falls back on `__deepcopy__` anyways, so I suggest adding a `__deepcopy__` method and optionally have `clone` point to it.

Also I would like to have support for pickling if possible. It is necessary when using python's concurrency as provided by [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) for example. Although https://github.com/jmschrei/pomegranate/issues/108 adds support for multithreading inside the fitting procedure, I think having the ability to fit distinct models in parallel would be a nice addition.

And while I am on the subject, do you support scikit-learn's [GridSearchCV](http://scikit-learn.org/stable/modules/generated/sklearn.grid_search.GridSearchCV.html#sklearn.grid_search.GridSearchCV) ?

Below is a sample code illustrating what I mean (slightly larger than necessary, sorry):

```python
#!/usr/bin/env python3

"""This is the (possibly) usual workflow for fitting an HMM with GMM emissions
"""

from concurrent.futures import ProcessPoolExecutor
from itertools import repeat

import numpy as np

from pomegranate.base import State
from pomegranate.gmm import GeneralMixtureModel
from pomegranate.distributions import IndependentComponentsDistribution, \
    DiscreteDistribution, NormalDistribution
from pomegranate.hmm import HiddenMarkovModel


# GMM helper stuff ------------------------------------------------------------

def check_gmm(gmm):
    """Check wether the parameters of a GMM are not Inf or NaN.
    """
    for component in gmm.distributions:
        for gaussian in component.distributions:
            if not np.all(np.invert(np.isnan(gaussian.parameters))):
                return False
    return True
    

def make_train_gmm(n_components, x):
    """Build and fit a Gaussian mixture model.

    .. note:: Gaussians are restricted to diagonal covariance matrix.
    """

    gmm = GeneralMixtureModel([
        IndependentComponentsDistribution([
            NormalDistribution(1, 1) for _ in range(x.shape[1])])
        for _ in range(n_components)])
    gmm.fit(x)

    # EM seldom fails to converge, just restart the process if it happens
    return gmm if check_gmm(gmm) else make_train_gmm(n_components, x)
    
    
def secs2mat(sequences, n):
[...]

PR #154, opened 2016-07-27
Add support for serialization in GMM and HMM

pixelou:
This PR follows up issue https://github.com/jmschrei/pomegranate/issues/153

It adds the methods `__getstate__` and `__getstate__` to GeneralMixtureModel and HiddenMarkovModel to make them 'picklable'.

I had to modify the automatic name generation of State to make sure we always have unique state names (unless overriden by user). Otherwise, deserialization fails on name collision. This fix relies on the uuid library which is part of python standard library.
There are two very simple regression tests to check that pickle doesn't crash. I didn't test if the reloaded value is identical to the original one because json tests should be sufficient. And I have no clue how to properly compare the objects :-) .

### DIFF
--- a/pomegranate/MarkovChain.pyx
+++ b/pomegranate/MarkovChain.pyx
@@ -48,13 +48,16 @@ cdef class MarkovChain(object):
 		self.k = len(distributions) - 1
 		self.distributions = distributions
 
+	def __reduce__(self):
+		return self.__class__, (self.distributions,)
+
 	def log_probability(self, sequence):
 		"""Calculate the log probability of the sequence under the model.
 
 		This calculates the first slices of increasing size under the
 		corresponding first few components of the model until size k is reached,
-		at which all slices are evaluated under the final component. 
-		
+		at which all slices are evaluated under the final component.
+
 		Parameters
 		----------
 		sequence : array-like
@@ -124,12 +127,12 @@ cdef class MarkovChain(object):
 			a sequence of variable length
 
 		weights : array-like, shape (n_samples,), optional
-			The initial weights of each sample. If nothing is passed in then 
+			The initial weights of each sample. If nothing is passed in then
 			each sample is assumed to be the same weight. Default is None.
 
 		inertia : double, optional
 			The weight of the previous parameters of the model. The new
-			parameters will roughly be old_param*inertia + new_param*(1-inertia), 
+			parameters will roughly be old_param*inertia + new_param*(1-inertia),
 			so an inertia of 0 means ignore the old parameters, whereas an
 			inertia of 1 means ignore the new parameters. Default is 0.0.
 
@@ -154,7 +157,7 @@ cdef class MarkovChain(object):
 			a sequence of variable length
 
 		weights : array-like, shape (n_samples,), optional
-			The initial weights of each sample. If nothing is passed in then 
+			The initial weights of each sample. If nothing is passed in then
 			each sample is assumed to be the same weight. Default is None.
 
 		Returns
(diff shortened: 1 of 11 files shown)


pixelou:
Ok, there is a small mistake in the tests, but that is not all. It seems python 2.7 is not as good as 3.* for pickling. Even simpler types such as DiscreteDistribution cannot be converted, I have to investigate this a bit deeper.

jmschrei:
Thanks for the contribution, and for looking into it for me! I'm eager to see what you find. Keep in mind that all objects are cython extension objects with pointers as private attributes. I don't know if this makes things more difficult, but I thought I'd let you know. In this sense, DiscreteDistribution is in fact one of the more complicated objects because even though the python side is just a dictionary, the cython side is several pointer arrays used for speedy training.

pixelou:
After further reading, I found that extension classes need to provide the `__reduce__` method instead of `__getstate__` and `__setstate__`  [http://ask.sagemath.org/question/8376/pickling-extension-classes/](http://ask.sagemath.org/question/8376/pickling-extension-classes/). `__reduce__` requires to add a function at the module level to process the reduced representation. Unfortunately, static methods or class methods are not allowed.  Is it ok if I add a `_deserialize` function in the modules?

jmschrei:
Sure, go ahead. I may end up renaming or restructuring it in the future to be more consistent with the API of the entire module, but it's more important to get something working right now. Thanks again for your efforts.

pixelou:
Modifications don't break the library anymore. There are still a few more models to do.

pixelou:
I'm going to leave the finite state machines and the factor graphs untouched because I am not familiar with these two and there are no serialization tests to build on.

Using serialization, it should be easy to propose a default `__deepcopy__` method in `Distribution` and `Model` to inherit from:

```python
    def __deepcopy__(self, memo):
        return pickle.loads(pickle.dumps(self))
```

That should in turn allow the use of pomegranate models in scikit-learn tools such as GridSearchCV.

jmschrei:
Thanks for this great work. I'll review it ASAP

jmschrei:
This looks great. Thanks for adding in unit tests as well. Would you mind squashing the commits to be a single one? Also, does this make the code compatible with GridSearchCV, or do I need to add the `__deepcopy__` method? 

HMM needs the `__getstate__` and `__setstate__` methods because converting it to/from JSON is more complicated than just using the constructor, right? The other models can just pass a list of parameters to feed into the constructor, but HMMs have structure?

pixelou:
I will look into git doc for commits squashing.

Unfortunately, I don't think supporting  GridSearchCV cleanly is going to be completely straightforward. The documentation on the estimator interface ([here](http://scikit-learn.org/stable/developers/contributing.html#rolling-your-own-estimator)) mentions the `get_params` and `set_params` that also need to be implemented. As I read the doc more carefuly, it's probably necessary to implement `clone` explicitely as well. The standard `__deepcopy__` is more like a fallback. To conclude on this point, I imagine the simplest solution is to implement the standard `__deepcopy__` using the simple solution based on pickle, and have the `clone` as an alias to comply with the estimator interface. I think it can be done by just modifying `Model` and `Distribution` base classes.

About the HMM: Actually, I didn't use Json methods for any of the distributions and models because JSON serialization uses class methods and pickle serialization relies on instance methods. The way I understand the pickle documentation ( and StackOverflow ;-) ) it's easier to simply have the constructor rebuild the object in most cases. If the constructor is not enough, as it is for HMMs, `__setstate__` (and its friend `__getstate__`) is the standard way to do the extra work.

jmschrei:
LGTM, I'm gonna merge it.

PR #171, opened 2016-10-04
add entries to gitignore

nelson-liu:
I added some entries to the gitignore specific to mac / emacs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 *.pyc
 *.c
 .ipynb_checkpoints
+*~
+.DS_Store
+build


jmschrei:
Does anyone even use those?

nelson-liu:
yup, I do ;) not sure why the `build` folder is generated on my machine, though...I'm presuming it isn't on yours since it wasn't in the gitignore before?

jmschrei:
I just never added it to the project. This is a good addition though. I'll merge this as soon as the tests finish running.

PR #172, opened 2016-10-04
add list of dependencies to readme

nelson-liu:
I added a brief list of the dependencies to the readme; took me awhile to get everything that was needed since errors that they're missing aren't necessarily thrown until test-time.


nelson-liu:
also is there a reason why you're pip installing a particular joblib version in `travis.yml` instead of just doing `conda install joblib`?

jmschrei:
When I first got travis working, the bleeding edge of joblib supported a feature that I needed, but the default pip/conda installation gave the wrong version. This can be updated now. Good catch!

nelson-liu:
ok, i'll leave it to you to make a fix 😄  thanks!

PR #178, opened 2016-10-09
unpin joblib version and install with conda

nelson-liu:
Noticed that you never got around to doing it, this should fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,9 @@ install:
   - conda info -a
 
   # Replace dep1 dep2 ... with your dependencies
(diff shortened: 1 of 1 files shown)


jmschrei:
Looks like Python 3.3 is messed up?

nelson-liu:
Seems like the with syntax was introduced in joblib 0.9, and conda installed 0.8.4 for some reason...see https://github.com/joblib/joblib/issues/241

nelson-liu:
I guess better to use pip (https://pythonhosted.org/joblib/installing.html) will fix tomorrow

PR #187, opened 2016-10-12
beautify kmeans and MarkovChain

pixelou:


jmschrei:
Great. All of the files should have a single line at the end though (I might've missed this for base), can you add that and then I'll merge?

pixelou:
Forgot to check that indeed. Fixed now.

jmschrei:
LGTM

jmschrei:
Thanks for the help with this. I think the best next place would be gmm and Naive Bayes, as I don't think I'll be touching those for a while.

PR #193, opened 2016-10-19
beautification

pixelou:
beautified gmm and kmeans + small fix of gmm.__reduce__ with regression test.

### DIFF
--- a/pomegranate/base.pyx
+++ b/pomegranate/base.pyx
@@ -35,8 +35,24 @@ cdef class Model(object):
 	def set_params(self, state):
 		self.__setstate__(state)
 
-	def to_json(self):
-		"""Serialize this object into JSON format."""
+	def to_json(self, separators=(',', ' : '), indent=4):
+		"""Serialize the model to a JSON.
+
+		Parameters
+		----------
+		separators : tuple, optional
+			The two separaters to pass to the json.dumps function for formatting.
+			Default is (',', ' : ').
+
+		indent : int, optional
+			The indentation to use at each level. Passed to json.dumps for
+			formatting. Default is 4.
+
+		Returns
+		-------
+		json : str
+			A properly formatted JSON object.
+		"""
 		raise NotImplementedError
 
 	@classmethod
--- a/pomegranate/base.pyx
+++ b/pomegranate/base.pyx
@@ -41,7 +41,8 @@ cdef class Model(object):
 		Parameters
 		----------
 		separators : tuple, optional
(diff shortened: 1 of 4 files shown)


pixelou:
Not much to change this time.
By the way, is it me or gmm ignores weights when callable distributions are given?

pomegranate/gmm.pyx
```diff
@@ -305,9 +285,12 @@ cdef class GeneralMixtureModel( Model ):
 		cdef double log_probability_sum = NEGINF
 		cdef double log_probability
 
-		for i in range( self.n ):
-			log_probability = ( <Model> self.distributions_ptr[i] )._log_probability(X) + self.weights_ptr[i]
-			log_probability_sum = pair_lse( log_probability_sum, log_probability )
+		for i in xrange(self.n):
```
jmschrei: There's no need to use xrange in cython code because it'll optimize range from a python loop to a C style loop.

pomegranate/base.pyx
```diff
@@ -35,8 +35,24 @@ cdef class Model(object):
 	def set_params(self, state):
 		self.__setstate__(state)
 
-	def to_json(self):
-		"""Serialize this object into JSON format."""
+	def to_json(self, separators=(',', ' : '), indent=4):
```
jmschrei: Are you using 6 spaces? We should be using 4

pomegranate/gmm.pyx
```diff
@@ -154,64 +128,71 @@ cdef class GeneralMixtureModel( Model ):
 	cdef double* summaries_ptr
 	cdef dict keymap
 	cdef int n
-	cdef public int hmm
+	cdef public bint is_hmm_
```
jmschrei: I had been putting changing this off for a bit, but it should probably be "is_vl_" for variable length. The issue isn't if it's a HMM or not, it's that the input is variable length and so needs to be handled slightly differently. Can you make that change?

jmschrei:
Thanks for the work. I've made some comments.

pomegranate/base.pyx
pixelou: Sorry I do not understand what you mean.

jmschrei: It looks like your indentation is set at 6 spaces not 4

pixelou: No it is not. What makes you think so?

jmschrei:
Great, thanks for the additions. The next one to do should be naive bayes, which is likely very similar to GMM.

Issue #188, opened 2016-10-13
Structural Optimization of the HMM topology


Narsail:
Hello Jacob,

do you have any plans to implement a separate or joint (merged fitting and topology training methods) topology training for the HMMs? There is the possibility to further increase the precision of the models through incorporating splitting and pruning of states.

jmschrei:
I am looking to try to add a structure learning method at some point in the next few months, but I don't currently have plans to add in a merged fitting and structure modification step. Do you have a good paper for the procedure you're looking for? It'd help me figure out approximately how much time adding it would take.

Narsail:
Not yet. I'm looking into this matter myself but it is hard to find complete literature about this problem. Hopefully i can tell you more in the next few days.

jmschrei:
Great. If you can find a structure + parameter learning algorithm directly from data that'd be the best for pomegranate. Have you seen this paper? http://repository.cmu.edu/cgi/viewcontent.cgi?article=2324&context=compsci

Narsail:
At the moment i'm looking into: http://www.sciencedirect.com/science/article/pii/S0031320399002265 and i'm planning to lend tomorrow the following book: http://www.springer.com/us/book/9783662533000 from my local library. Thanks for the paper!

Narsail:
I don't have a clear course yet but what i will do is to implement the pruning and splitting of states for the HMM. Hopefully i can provide you a pull request soon.

Narsail:
So i added the pruning of transitions as well as the splitting of a existing state in the pull request #194.

For the state splitting it is necessary to split the distribution as well. At the moment there are only implementations for the Normal and the IndependentComponentDistribution (because those are in use in my project^^)

jmschrei:
Thanks for taking the time to try to incorporate this work into pomegranate. Unfortunately, I got side tracked in the last year and this never made it into pomegranate. If you'd like to propose a structure learning addition, it should follow the format of `HiddenMarkovModel.from_samples`, which learns a dense graph. Please open a new issue/PR if you do, and I'll try to provide guidance!

PR #194, opened 2016-10-20
Foundation for Structural Optimization of the HMM topology

Narsail:
See #188

### DIFF
--- a/pomegranate/distributions.pyx
+++ b/pomegranate/distributions.pyx
@@ -664,12 +664,12 @@ cdef class NormalDistribution( Distribution ):
 		mu = self.summaries[1] / self.summaries[0]
 
 		''' The Original Formula var = x2 / x0 - x1^2 / x0^2 could lead to ZeroDivision Errors
-		through extreme small valus in x0 (i.e. 1.7^-300) which resulted through the ** 2 in the
-		value 0. '''
+		through extreme small values in x0 (i.e. 1.7^-300) which resulted in the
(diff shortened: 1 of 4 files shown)


jmschrei:
What is the difference between splitting a state and turning it into a mixture of the two distributions with priors corresponding to the relative likelihood of going to each state?

Narsail:
The idea is to give you a tool to optimize the structure of the HMM. I.e. in speech recognition the pronunciation of the letter "i" can be different from person to person. The Distribution mixture for every state will stay the same in kind and numbers (in speech recognition you work with a fixed number of features which will be represented by the Distribution of the state). So if you want to represent different pronunciations in one HMM you need to refine the structure and therefore the number of states and edges. The splitting of a state gives you a simple approach to extend that network. 

Btw. pulled the forked repository via pip into my project but can't find my methods (i pulled directly the commit). Do you know why this might happen?

jmschrei:
I guess what I'm asking is what is the difference between splitting the state into two with the same in-out states and just replacing the distribution of that state with a mixture of the two distributions, with the prior probability over each distribution representing the difference in probabilities that the edges would've represented.

Narsail:
In a simple left to right structure it makes sense to create a mixture distribution if you only want to split one state but if you want to build a new path through the hmm only working with the priors and mixture models won't work. So considering the splitting step isolated it can be done through mixture distributions. If you want to train a model, i.e. do multiple splits and then work with fitting and pruning to build a new structure, you need to split states.

PR #207, opened 2016-11-28
Support for multivariate discrete

jmschrei:


jmschrei:
This PR add support for multivariate discrete emissions in many models, mainly hidden Markov models. It also cleans up the code a bit to add nicer input checks for HMMs, adds a new error message for HMM input and two new unit tests to cover these cases.

Issue #214, opened 2017-01-17
BayesianNetwork.predict modify item parameter


DeoLeung:
```python
model = BayesianNetwork('example')
...
observations = [None, a, b]
predictions = model.predict([observations])
```

the observations will be modified after predict https://github.com/jmschrei/pomegranate/blob/master/pomegranate/BayesianNetwork.pyx#L543

is it intentional?

DeoLeung:
and it doesn't support `numpy.ndarray` as input as the docstring says

https://github.com/jmschrei/pomegranate/blob/master/pomegranate/BayesianNetwork.pyx#L515

https://github.com/jmschrei/pomegranate/blob/master/pomegranate/BayesianNetwork.pyx#L537

```sh
In [59]: k_ = model.predict(k)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-59-9063534cb611> in <module>()
----> 1 k_ = model.predict(k)

pomegranate/BayesianNetwork.pyx in pomegranate.BayesianNetwork.BayesianNetwork.predict (pomegranate/BayesianNetwork.c:9450)()

pomegranate/BayesianNetwork.pyx in pomegranate.BayesianNetwork.BayesianNetwork.predict_proba (pomegranate/BayesianNetwork.c:8012)()

ValueError: State '4609' does not have key 'nan'

In [60]: type(k)
Out[60]: numpy.ndarray
```

k contains np.nan

jmschrei:
1) There is no reason that the observations array has to be modified. I put that in there for expediency likely. Are you finding that this was a bad design choice? It can be changed. 

2) Good catch. It's not that it doesn't support numpy arrays, it's that it's not counting numpy.nan as a "missing value" as it would need to in this line: https://github.com/jmschrei/pomegranate/blob/master/pomegranate/BayesianNetwork.pyx#L537

I can look into making both those changes soon.

PR #216, opened 2017-01-18
BayesianNetwork.predict change:

DeoLeung:
1. it doesn't modify input observations now and always return a numpy.ndarray
2. it identify np.nan as missing value correctly

#214


jmschrei:
Great, thanks for the contribution!

PR #217, opened 2017-01-23
Fix generating exact Bayesian networks

jagub2:
Hello,

Today I tried creating naive Bayesian networks (all nodes have one edge, to the only root - label) in pomegranate, using `exact` algorithm and passing constraint_graph, but I couldn't due to unhandled exceptions. Upon investigation, I found out that tuples weren't added correctly.

I'm not quite sure about next part, about deleting iterating over children.

@jmschrei - can you please have a look into this?

Regards,
J.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -1075,7 +1075,7 @@ cpdef discrete_exact_with_constraints(numpy.ndarray X, numpy.ndarray weights,
 	structure = [None for i in range(d)]
 
 	for parent, child in constraint_graph.edges():
-		parent_sets[child] += parent
+		parent_sets[child] += (parent,)
 		if child == parent:
 			cycle[indices[child]] = 1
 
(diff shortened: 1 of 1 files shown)


jmschrei:
Howdy

Can you send me some minimal code reproducing the issue so I can better understand exactly what you mean? The values stored in constraint graphs are supposed to be a list of indices, not just a single number. The list can have a single value. Does changing the value to be a list fix the issue?

jagub2:
Hello,

I'm passing `networkx.DiGraph`. Code reproducing the issue: https://gist.github.com/jagub2/63c339e352886468e489c87c2c717d2f; dataset: http://mldata.org/repository/data/viewslug/uci-20070111-breast-cancer/

Error:

```Traceback (most recent call last):
(...)
  File "pomegranate\BayesianNetwork.pyx", line 785, in pomegranate.BayesianNetwork.BayesianNetwork.from_samples (pomegranate\BayesianNetwork.c:13899)
  File "pomegranate\BayesianNetwork.pyx", line 1074, in pomegranate.BayesianNetwork.discrete_exact_with_constraints (pomegranate\BayesianNetwork.c:18101)
TypeError: can only concatenate tuple (not "int") to tuple
```

Changing source of edge to list or tuple doesn't help.

Regards,
J.

jmschrei:
Try changing https://gist.github.com/jagub2/63c339e352886468e489c87c2c717d2f#file-test-py-L25 and L26 to the following:

```
        for i in range(1, len(self.state_names)):
            graph.add_edge((0,), (i,))
```

jagub2:
Hello,
I changed it and now it works, thank you. Looks like no change in `pomegranate` code is needed. 
I got confused by `constraint_graph` description in documentation, so I looked to `networkx` documentation, and reading the docs gave me this idea.
Thank you for checking this.

jmschrei:
No problem. I can see how you got confused. I will add a patch soon which allows you to pass in integers as well, since that seems a reasonable thing that someone would want to do.

PR #222, opened 2017-02-08
Some updates

complyue:


jmschrei:
Hi, thanks for the update. What's the purpose of adding the `#!python` marker? No one should be running those scripts directly, and the python interpreter would likely crash anyway.

complyue:
wrt `#!python`, it's for better-than-nothing syntax highlighting in IntelliJ's PyCharm IDE community edition, as PyCharm professional edition supports Cython, their community edition doesn't. And it was an occasional finding at viewing `pomegranate/distributions.pyx` in the IDE, where the source is highlighted as python syntax, while rest pyx files  were not. Then after some investigation, I found it's due to the shebang, since distributions.pyx already has it, I'd think it's worthy to add to other files for better looking in PyCharm.
And seems you are not using PyCharm, if you opt these shebangs out, I'd concur.

jmschrei:
Oh that's interesting. I'm okay with adding it in, but can you either (1) change all tests to use numpy.isclose or (2) get rid of the few you've added? I don't want there to be a divergence in the testing.

complyue:
okay, I'll play with the tests wrt all `== 0` checks when I got time, might be some later.

pomegranate/base.pyx
```diff
@@ -137,6 +138,29 @@ cdef class Model(object):
 		"""
 		raise NotImplementedError
 
+	def plot( self, n=1000, **kwargs ):
```
jmschrei: This isn't going to work on multivariate distributions or models.

complyue: oops, didn't think of that, I'm plotting some dists for one of my PoCs, and for multivariate dists, overlay plotting of all variables seems satisfactory to me:

image

with updated `plot` code like:
```
	sampled_data = numpy.array(self.sample(n))
	for i in range(sampled_data.shape[-1]):
		plt.hist( sampled_data[..., i], **kwargs )
```

Do you think this implementation makes sense?

pomegranate/utils.pyx
```diff
@@ -118,8 +118,9 @@ cdef double gamma(double x) nogil:
 	cdef double den, num, result, z, y
 	cdef int i, n, arg_was_less_than_one
 
-	if x == 0.0:
-		return INF
+	with gil:
```
complyue: performance impacts by reacquiring GIL here is not clear yet

complyue: just realised it's comparing to zero, so `-1e-9 < x < 1e-9` might be an acceptable option.
thoughts?

pomegranate/base.pyx
jmschrei: I can't think of a better implementation right now, but that method has some issues. In particular it assumes a diagonal covariance matrix which can be misleading. I suppose it's better than not having anything at all right now, as long as it's properly documented to show this limitation.

pomegranate/utils.pyx
jmschrei: That restriction seems fine. In general it is not okay to reacquire the GIL on low level functions like those in utils or the distributions because that will easily cascade into thousands of GIL acquiring/releasing which will mess with the multithreading. Doesn't the case below (x < 0.001) cover this case though?

pomegranate/distributions.pyx
```diff
@@ -49,7 +49,7 @@ def log(value):
 	if isinstance( value, numpy.ndarray ):
 		to_return = numpy.zeros(( value.shape ))
 		to_return[ value > 0 ] = numpy.log( value[ value > 0 ] )
-		to_return[ value == 0 ] = NEGINF
+		to_return[ numpy.isclose(value, 0) ] = NEGINF
```
jmschrei: Why is this necessary?

pomegranate/base.pyx
```diff
@@ -454,7 +484,8 @@ cdef class State(object):
 		weight = d['weight']
 
 		c = d['distribution']['class']
-		dist = eval(c).from_json( json.dumps( d['distribution'] ) )
+		import pomegranate as _global_
+		dist = eval(c, _global_.__dict__).from_json( json.dumps( d['distribution'] ) )
```
jmschrei: What does this do?

pomegranate/base.pyx
jmschrei: Adding a plot function should be a separate PR. I would prefer not to merge several disparate features in a single PR unless necessary.

pomegranate/distributions.pyx
complyue: this is covered by `replace floating point == 0 tests with isclose while retain all tests passing` goal, more justification may be applied.

pomegranate/base.pyx
complyue: this fixed a bug in `hmm.from_json()` when gmm is used in the `.to_json()` saved model.

complyue:
@jmschrei I do agree that all those changes of my commits should be categorised into separate PRs, but I'm still figuring out how to achieved that, it seems all my subsequent commits will automatically go into the 1st PR which only some minor updates I meant. I admit that I'm not familiar with how Git/GitHub works, can I get some quick directions from you? I am supportive in any efforts to make this great project in organised thus high quality.

PR #249, opened 2017-04-28
adding pandas.DataFrame option to BayesianNetwork.predict

### LINKED ISSUE #248: pandas compatibility
@jmschrei et al.
Is there any chance to make wrappers to make pomegranate pandas friendly? 
I.e, I would like to feed dataframes and get some of the predictions in dataframes.

I am currently running on some scripts and look forward to your feedback (be gentle, please, I'm a newbie)

elzurdo:
Looking to resolve #248 

Please be gentle ... I'm a newbie. I love this module and would like to add pandas options. I hope that is fine. This is my first attempt. 

The regular call for prediction still works:
`model.predict(df_test.values)` yields  
```
array([[0.0, 4.0, 6.0, 36, 3.0, 5.0, 0.0, 'Male'],
       [4.0, 2.0, 3.0, 48, 0.0, 7.0, 0.0, 'Female'],
       [0.0, 8.0, 9.0, 19, 2.0, 4.0, 9.0, 'Male'],
       [0.0, 8.0, 6.0, 19, 2.0, 4.0, 0.0, 'Male']], dtype=object)
```

And now with the actual `pd.DataFrame` it preserves the original indexes and columns:
```
            col_0 col_1 col_2 col_3 col_4 col_5 col_6   col_7
uid                                                          
33000000067     0     4     6    36     3     5     0    Male
33000000593     4     2     3    48     0     7     0  Female
33000000715     0     8     9    19     2     4     9    Male
33000000718     0     8     6    19     2     4     0    Male
```

I also updated the documentation, 
and made sure to include case in which the user does not have pandas installed.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -38,6 +38,11 @@ try:
 except ImportError:
 	pygraphviz = None
 
+try:
+	import pandas as pd
+except:
+	pd = None
+
 if sys.version_info[0] > 2:
 	# Set up for Python 3
 	xrange = range
@@ -531,7 +536,7 @@ cdef class BayesianNetwork( GraphModel ):
 
 		Parameters
 		----------
-		items : array-like, shape (n_samples, n_nodes)
+		items : array-like, shape (n_samples, n_nodes), pandas.DataFrame
 			Data matrix to impute. Missing values must be either None (if lists)
 			or np.nan (if numpy.ndarray). Will fill in these values with the
 			maximally likely ones.
@@ -542,12 +547,21 @@ cdef class BayesianNetwork( GraphModel ):
 
 		Returns
 		-------
-		items : numpy.ndarray, shape (n_samples, n_nodes)
+		items : numpy.ndarray, shape (n_samples, n_nodes), pandas.DataFrame
 			This is the data matrix with the missing values imputed.
 		"""
 
 		if self.d == 0:
 			raise ValueError("must bake model before using impute")
+
+
+		if pd:
+			if isinstance(items, pd.DataFrame):
+				items_indexes = items.index
+				items_columns = items.columns
+				items = items.values.copy()
+			else:
+				items_indexes = None
 
 		imputations = numpy.copy(items)
 		for i in range(len(items)):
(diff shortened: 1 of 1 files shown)


jmschrei:
Thanks for the contribution. In general, it looks good. However, the bigger question for me is an API one. Specifically, it seems incomplete to only have one method in one model support data frames, but it's unclear how data frames might be supported for other models. Do you have any suggestions for how to extend pandas support to other models, so that they all accept/return pandas dataframes?

elzurdo:
Hi @jmschrei,
I fully understand your concern. 

In brief - 

this is something that I would have to figure out by learning more about the package, and I'm happy to contribute. For the most part arrays and dictionaries are fine, but in some (particular) cases I think that entering a dataframe and getting one back might be beneficial for tracking indexes and column names. 

In detail - 

* `pd.DataFrame`s can be naturally used  in commands like `numpy.array(X)` (as in `BayesianNetwork.from_samples`) without clash. 

* The downside of the striping of indexes and columns means that anytime one uses the model they have to be careful with the ordering. Yes, dictionaries are useful for this, but `pd.DataFrame`s, in some cases, as the one in this example simplify this tracking, too.

I am particularly interested in testing how the BN compares with Decision Trees in some aspects, e.g examining performance on one target variable. Dataframes as input and output are useful for this. To this extent I am glad to develop useful modules. 

E.g I created a `predict_proba_df`, which I am glad to make a pull request for, which takes in a `pd.DataFrame` where the target column is Nulled and returns a `pd.DataFrame` with the original entry indexes and each column is a class in the categorical of interest where the values are the probabilities. 

E.g, 
```
                Male Female 
uid                                                          
33000000067     0.64     0.36     
33000000593     0.44     0.56    
33000000715     0.22     0.78     
33000000718     0.15    0.85     
```

If you think that this is not of interest in general, I can appreciate that and will leave this particularity to myself.

jmschrei:
Sorry for the late response. Essentially, I would be okay with adding in pandas support for all main models in a PR, but not for a single method for a single model. It should be fairly easy to add in to naive Bayes, Bayes classifier, mixtures, and Bayesian networks for the predict step. I don't see a clear way to support the `predict_proba` method so it's probably okay to ignore those for now. Potentially this should also be added for the fit step for those methods.

PR #260, opened 2017-05-15
[MRG] readthedocs update

jmschrei:
Documentation update for recent API changes.

### DIFF
--- a/docs/HiddenMarkovModel.rst
+++ b/docs/HiddenMarkovModel.rst
@@ -6,63 +6,80 @@ Hidden Markov Models
 - `IPython Notebook Tutorial <https://github.com/jmschrei/pomegranate/blob/master/tutorials/Tutorial_3_Hidden_Markov_Models.ipynb>`_
 - `IPython Notebook Sequence Alignment Tutorial <http://nbviewer.ipython.org/github/jmschrei/yahmm/blob/master/examples/Global%20Sequence%20Alignment.ipynb>`_
 
-`Hidden Markov models <http://en.wikipedia.org/wiki/Hidden_Markov_model>`_ (HMMs) are a form of structured model in which a sequence of observations are labelled according to the hidden state they belong. A strength of HMMs is their ability to analyze variable length sequences whereas other models require a static set of features. This makes them extensively used in the fields of natural language processing and bioinformatics where data is routinely variable length sequences. They can be thought of as a structured form of General Mixture Models. 
+`Hidden Markov models <http://en.wikipedia.org/wiki/Hidden_Markov_model>`_ (HMMs) are a structured probabilistic model that forms a probability distribution of sequences, as opposed to individual symbols. It is similar to a Bayesian network in that it has a directed graphical structure where nodes represent probability distributions, but unlike Bayesian networks in that the edges represent transitions and encode transition probabilities, whereas in Bayesian networks edges encode dependence statements. A HMM can be thought of as a general mixture model plus a transition matrix, where each component in the general Mixture model corresponds to a node in the hidden Markov model, and the transition matrix informs the probability that adjacent symbols in the sequence transition from being generated from one component to another. A strength of HMMs is that they can model variable length sequences whereas other models typically require a fixed feature set. They are extensively used in the fields of natural language processing to model speech, bioinformatics to model biosequences, and robotics to model movement.  
+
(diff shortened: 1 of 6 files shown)


docs/HiddenMarkovModel.rst
```diff
@@ -6,63 +6,80 @@ Hidden Markov Models
...
+
+The second way to initialize models is to use the ``from_samples`` class method. The call is identical to initializing a mixture model.
+
+.. code-block:: python
+
+	from pomegranate import *
+	model = HiddenMarkovModel.from_samples(NormalDistribution, n_components=5, X=X)
+
+Much like a mixture model, all arguments present in the ``fit`` step can also be passed in to this method.
+
```
jkleckner: Nice improvements.  Would it be possible to add some comments here about how covariances are handled with the different ways of initializing as well as if further training is performed?

jmschrei: In all methods, when you initialize a multivariate gaussian using kmeans you calculate the means and covariances of each cluster to get the initial parameters. Do you think that should be explicitly written in?

jkleckner: Sorry for such a cryptic comment.  When I read more carefully the Bayes comment about IndependentComponentsDistribution vs MultivariateGaussianDistribution I see it doesn't apply to the question.

I was motivated by a mention in an email thread that you could initialize a model parametrically with starting multivariatedists/covars/transitions as a starting point and then adjust the model from that starting point using the data.  Is that possible, does it make sense, and, if so, where would it fit into the documentation?

jmschrei:
@jkleckner your comments seem to have been masked by my recent update where I go ahead and explain it explicitly. The idea is that you run k-means on the data initially, and then initialize your distributions on the resulting clusters. If you distribution is a multivariate Gaussian distribution, you initialize both mean and covariances on the clusters. If it's an independent components distribution, each distribution is initialized separately for each dimension, but still on the resulting clusters.

docs/HiddenMarkovModel.rst
```diff
@@ -6,63 +6,80 @@ Hidden Markov Models
...
 
-Initialization
---------------
-
-Transitioning from YAHMM to pomegranate is simple because the only change is that the ``Model`` class is now ``HiddenMarkovModel``. You can port your code over by either changing ``Model`` to ``HiddenMarkovModel`` or changing your imports at the top of the file as follows:
+The HMM implementation in pomegranate is based off of the implementation in its predecessor, Yet Another Hidden Markov Model (YAHMM). To convert a script that used YAHMM to a script using pomegranate, you only need to change calls to the ``Model`` class to call ``HiddenMarkovModel``. For example, a script that previously looked lik the folllowing:
```
jkleckner: Note typos: `lik => like` and `folllowing => following`

docs/NaiveBayes.rst
```diff
@@ -1,86 +1,111 @@
...
-
-An IPython notebook example demonstrating a Naive Bayes classifier using multivariate distributions can be `found here <https://github.com/jmschrei/pomegranate/blob/master/examples/naivebayes_multivariate_male_female.ipynb>`_.
+Bayes classifiers are simple probabilistic classification models based off of Bayes throerom. See the above tutorial for a full primer on how they work, and what the distinction betwene a naive Bayes classifier and a Bayes classifier is. Essentially, each class is modeled by a probability distribution and classifications are made according to what distribution fits the data the best. They are a supervised version of general mixture models, in that the ``predict``, ``predict_proba``, and ``predict_log_proba`` methods return the same values for the same underlying distributions, but that instead of using expectation-maximization to fit to new data they can use the provided labels directly.
```
jkleckner: Typo `betwene => between`

jkleckner:
@jmschrei Yes the newer text explains that mechanism.  The force pushes are making some changes disappear as well.  I guess I'm asking about a way to seed the partition and parameters independent of kmeans and then letting the fit process improve and adjust the model from that starting point.  Is there a way to do that?  I can't find the email that came over the transom but I thought I saw something that hinted at how to do that.

jmschrei:
If you'd like to initialize the parameters yourself, the first initialization method for each of these objects is the way to go. For HMMs, you can manually build the graph with initial distributions, and then call the `fit` method. For mixture models, you can pass in a list of distributions that are the initial components, and then run `fit` as well. Does that solve the issue?

jkleckner:
@jmschrei "Does that solve the issue?"  Doh, of course, thanks.  It would be nice if the doc site could include the doc generated from the source code in addition to the rst.

jmschrei:
It should be fixed now.

jkleckner:
@jmschrei Awesome, thank you!

PR #264, opened 2017-05-27
Fix sampling from Conditional Probability Tables

MechCoder:
The Conditional Probability Tables method for sampling seems broken to me.

1. I get this error, TypeError: 'odict_keys' object does not support indexing
2. There is a  dict as a default value (which is mutated). This means the 2nd time I call `sample()`, the value of `parent_values` used is the one modified internally after calling `sample()` for the first time.

I also added a test.

### DIFF
--- a/pomegranate/distributions.pyx
+++ b/pomegranate/distributions.pyx
@@ -329,7 +329,7 @@ cdef class Distribution(Model):
 		elif 'Table' in d['name']:
 			parents = [Distribution.from_json(json.dumps(j)) for j in d['parents']]
 			table = []
-			
+
 			for row in d['table']:
 				table.append([])
 				for item in row:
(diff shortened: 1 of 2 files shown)


MechCoder:
I removed the imports from sklearn. (Most of the diff is due to whitespaces that atom automatically took care of).

Btw, I think you should enforce consistency in how `sample` is being called across different models. In some places, the first argument is the number of samples to generate and over here it is hardcoded to 1 and the first argument is `parent_vals`

pomegranate/distributions.pyx
```diff
@@ -2487,31 +2487,28 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 
 		self.keymap = OrderedDict(keymap)
 
-	def sample(self, parent_values={}):
+	def sample(self, parent_values=None):
```
jmschrei: Good catch, thanks!

jmschrei:
Thanks for the contribution! And yeah, I'm in the process of updating the API from single element to vectorized operations. pomegranate originally did every single element, such as calculating a single log probability or producing a single sample, but that's pretty slow when it comes to larger tasks. I'll look into updating it soon.

PR #265, opened 2017-05-31
docs: Fix the docstring of the HMM's `predict`

### LINKED ISSUE #245: HMM predict documentation not correct
The [predict documentation](https://github.com/rlcode/reinforcement-learning) says that it returns (logp, path). But when I ran my HMM code, I found that it only returned the state sequence, and only the numbers of the states at that, not even the entire object, like model.viterbi does.

predict returned `[3, 0, 1, 0, 2]` (where I have 3 states in my model)

mrshu:
* Previously, calling `predict` was the same as calling `viterbi` on a
  HMM object. This behaviour has since changed, but the docs stayed the
  same.

* This commit updates the inline docstring of the `predict` function of
  `HMM` in order to reflect the new situation.

* Fixes #245.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/pomegranate/hmm.pyx
+++ b/pomegranate/hmm.pyx
@@ -2271,12 +2271,8 @@ cdef class HiddenMarkovModel(GraphModel):
 
         Returns
         -------
-        logp : double
-            The log probability of the sequence under the Viterbi path
-
-        path : list of tuples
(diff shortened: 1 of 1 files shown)


jmschrei:
Howdy, thanks for the contribution!

Issue #175, opened 2016-10-04
Improve model fitting logging


jmschrei:
After each iteration in GMM/HMM training it should report the time it took to train that iteration. Calls to 'fit' should also return a "history" object like in keras which shows the dataset log probability at each iteration, time, and other relevant statistics.

mrshu:
@jmschrei if you do not mind I'd like to help you out with this, since I found `pomegranate` quite useful in my own work, and could actually benefit from the additions requested by this issue.

Since this is a pretty broad issue, I decided to tackle it in multiple passes. The first one being the first sentence of this issue:

> After each iteration in GMM/HMM training it should report the time it took to train that iteration.

I took it a small step further and also report the total time it took to train a given GMM/HMM model in #269.

My next steps here would be designing a history-like object, taking Keras as the main source of inspiration, per your description above. If you do not mind me asking, can you think of any other relevant statistics one would like to find in a "history" of training apart from time, improvement and log probability? 

Thanks!

PR #267, opened 2017-06-07
[MRG] Implement model persistence for FactorGraphs

MechCoder:

Issue #156, opened 2016-08-08
More uniform codestyle


nlgranger:
Hi,
I noticed a few opportunities for improvement in the codestyle of the project, here are a few:
- tab indentation -> 4 space indentation
- spaces inside parenthesis and before comme -> remove
- number of blank lines between functions or classes
- long lines -> break
- possibly others.

These are far from being a show breaker, but if it is okay, I am willing to work on these. However, it might create merge conflicts with pending work.

If you think this is useless, you can close.

jmschrei:
I don't mind someone going through and looking for improvements in the code style, but I don't know if I care about some of these things. Do you think it's worth your time to pursue this? What would the benefits be? It would likely have to wait until after the end of the month, as I am developing several aspects of the codebase for the next few weeks and don't want to interfere with your work.

pixelou:
Well, for the beauty of it is not enough of an argument ;-) ?  
The space thing is easy to fix and would comply with the default behaviour of most editors. The long lines is convenient for people with small screens or dual view as in diffs. The rest might be useful to enable linting and static checks. I will see if I can have access to pycharm's professial edition and run the linter on cython files.

Nevertheless, since this is indeed very low priority, you can close for now or tell me later when I can work on this.

jmschrei:
Okay. I appreciate the offer, regardless! I'll comment here again when we should discuss this again.

jmschrei:
Lets start discussing this now--I still have some features I need to add, but hopefully I'll be done by the time we come to a consensus.

1. What is the difference for you between tabs and spaces? My understanding is most editors now can arbitrarily change between them, and I personally prefer tabs.
2. I used to favor having a lot of spaces inside parenthesis, but I've decided I don't like it as much. Personally, it seems nicer to do something like `numpy.sum(X)`, but if you're doing a more complex things to add in spaces, such as `numpy.sum( numpy.abs(X) - numpy.mean(X) )`. The spaces, to me, help clarify the internal logic. Parameters to functions should not have spaces though, e.g. `def func( a, b )` should be changed to `def func(a, b)`.
3. I'm told it should be 2 lines between functions/classes. Is this right?
4. This should already be done. Lets take a look at cases where it is not done.

I'd like to hear your thoughts on this.

pixelou:
Hi there!

- Stackoverflow has rather long discussions about why they are equivalent, but I personnaly see a few advantages for spaces:
[...]

PR #269, opened 2017-06-11
hmm + gmm: Add timing information to `fit` output

mrshu:
* Add timing information to verbose output the `fit` function of GMM and
  HMM, as described in #175.

Signed-off-by: mr.Shu <mr@shu.io>


mrshu:
@jmschrei Sorry for so many commits -- I'll gladly squash them.

It turns out that one cannot really rely on the presence of `clock_gettime` on all of the considered platforms. In hindsight, I should have done it this way since the beginning -- getting the timing info does not necessarily have to be optimized in this case.

pomegranate/gmm.pyx
```diff
@@ -170,8 +170,8 @@ cdef class GeneralMixtureModel(BayesModel):
 			else:
 				improvement = log_probability_sum - last_log_probability_sum
 
-				time_spent = time_in_epoch_sec() - epoch_start_time
-				msg = "Improvement: {} in {:.2f}"
+				time_spent = time.time() - epoch_start_time
+				msg = "Improvement: {} in {:.2f}s"
```
jmschrei: Might as well just make this

```
if verbose:
    print "Improvement: {} in {:.2f}".format(improvement, time_spent)
``

**Comment by jmschrei** [comment]
Getting the current clock time is so inconsequential compared to the main computations that I don't think you should bother. Even if it took 100 ms, no one would really notice, and usually it takes less than 1ms.

**Comment by mrshu** [comment]
Thanks for your assessment @jmschrei, I agree with that.

The checks have passed, so this should be safe to merge now.

File: pomegranate/gmm.pyx [hunk 2]
```diff
@@ -167,17 +170,20 @@ cdef class GeneralMixtureModel(BayesModel):
 			else:
 				improvement = log_probability_sum - last_log_probability_sum
 
+				time_spent = time.time() - epoch_start_time
 				if verbose:
-					print("Improvement: {}".format(improvement))
+					print("Improvement: {} in {:.2f}s".format(improvement, time_spent))
```
mrshu: @jmschrei I am not sure if you meant it this way in your comment (I believe the `s` at the end was omitted), but the reason why I tried to split this into two lines was that currently the line is a bit too long when you set your `tabwidth` to 8.

However, since there are even longer lines in this file, this is acceptable in my mind too.

pomegranate/hmm.pyx
```diff
@@ -2486,8 +2490,10 @@ cdef class HiddenMarkovModel(GraphModel):
                     initial_log_probability_sum = log_probability_sum
                 else:
                     improvement = log_probability_sum - last_log_probability_sum
+                    time_spent = time.time() - epoch_start_time
                     if verbose:
-                        print("Training improvement: {}".format(improvement))
+                        print("Training improvement: {} in {:.2f}s".format(improvement,
+                                                                           time_spent))
```
mrshu: @jmschrei Just a quick note with regards to #156 -- the alignment of `improvement` and `time_spent` here is possible outright due to the use of spaces. If spaces were not used, the first part of the line (up until `print`) would have to be indented with tabs and from then on spaces could be used. This would make it quite difficult to put together.

mrshu:
@jmschrei The checks have passed -- I have contemplated adding some more tests for this. In the end, it does not seem worth it as their value would be quite negligible. I'll try to add more of them when implementing the other parts of #175.

Please let me know if there Is anything else you think should be done before merging this.

Thanks!

pomegranate/gmm.pyx
jmschrei: I suppose it doesn't really matter one way or another. This looks fine.

pomegranate/hmm.pyx
jmschrei: Ultimately it doesn't matter too much. It is obvious what is going on to people who want to develop.

jmschrei:
Is it possible to change the format to be as follows:

[iteration] Improvement: {improvement}, Time (s): {time}

starting with 1 as the first iteration, not 0. Thanks!

mrshu:
@jmschrei Sure, I'll do that right away!

pomegranate/hmm.pyx
mrshu: Agreed for sure.

pomegranate/gmm.pyx
```diff
@@ -167,17 +170,22 @@ cdef class GeneralMixtureModel(BayesModel):
 			else:
 				improvement = log_probability_sum - last_log_probability_sum
 
+				time_spent = time.time() - epoch_start_time
+				vals = dict(iteration=iteration, improvement=improvement, time=time_spent)
+				fmt = "[{iteration}] Improvement: {improvement} Time (s): {time:.2f}"
```
mrshu: @jmschrei I realize you may find this to be a bit too much, but I believe that it is better to create a variable this way than trying to fit it into the `print` line below, which would then overflow.

pomegranate/gmm.pyx
```diff
@@ -167,17 +170,22 @@ cdef class GeneralMixtureModel(BayesModel):
...
 			last_log_probability_sum = log_probability_sum
 
 		self.clear_summaries()
 
 		if verbose:
-			print("Total Improvement: {}".format(
-				last_log_probability_sum - initial_log_probability_sum))
+			total_imp = last_log_probability_sum - initial_log_probability_sum
+			total_time_spent = time.time() - training_start_time
+			print("Total Improvement: {}".format(total_imp))
```
mrshu: @jmschrei Do these prints look OK? I pretty much only guessed their format.

pomegranate/FactorGraph.pyx
```diff
@@ -395,7 +395,7 @@ cdef class FactorGraph( GraphModel ):
 			"states": [json.loads(state.to_json()) for state in self.states],
 			"edges": self.edges
 		}
-		return json.dumps(model, separators=separators, indent=indent)
+		return json.dumps(model, separators=separators, indent=indent, sort_keys=True)
```
mrshu: @jmschrei It turns out this was necessary in order for the JSON comparison test of factor graphs (https://github.com/jmschrei/pomegranate/pull/267/files#diff-6ba2f6007ae6a66d1e3941680c09a027R31) to work reliably.

jmschrei: Great catch, I just noticed a test failed on AppVeyor.

pomegranate/gmm.pyx
jmschrei: I don't mind merging it now and then revisiting the issue later. Better to have the feature then clean it up.

mrshu: Cool, thanks!

pomegranate/FactorGraph.pyx
mrshu: I am not sure if you would like me to send another patch specifically for this -- I am happy to do that, just let me know.

mrshu:
@jmschrei From my point of view, this PR is done so please let me know how you would like me to proceed here. I see two options:

1. Just merge this PR, which will both add the timing info and fix the FactorGraph test issue
2. Create a separate PR for the FactorGraph test issue, get that merged and then merge this PR.

Whichever option you choose, I am happy to help you with.

Thanks!

jmschrei:
I will be getting to this soon-- very busy recently! Thank you for your patience.

mrshu:
@jmschrei No worries at all -- I know very well how that is!

In the meantime, I added timing info to `BayesClassifier`, `NaiveBayes` and also `kmeans`, as their `fit` function was pretty similar to that of `gmm` and `hmm`. I believe this would be a nice addition, but I am happy to discuss it.

jmschrei:
Great catch, semi-supervised learning using those models is also iterative. I think those are the only ones, though. Bayes nets, Markov chains, and probability distributions just use MLE for parameter updates.

mrshu:
@jmschrei thanks, that is indeed what I gathered from grepping around through the repository.

PR #270, opened 2017-06-12
[WIP] Add GPU support for models

jmschrei:
This PR will add GPU support for all models, starting with multivariate gaussian distributions, GMMs, and HMMs, using the package `cupy`. Stay tuned!


pixelou:
Hi, just out of curiosity: have you considered using theano for the job?

jmschrei:
I did consider it, but ultimately decided that Theano was not the appropriate package for just doing matrix multiplications on a GPU. cupy provides a far simpler interface with possibility for cython-level integration, which would be very useful.

pixelou:
Ok.

PR #273, opened 2017-06-13
Remove trailing spaces tabs

jkleckner:
This patch set removes trailing spaces and tabs in one pass.  This can reduce extraneous noise with future pull requests.


docs/index.md~
```diff
@@ -33,7 +33,7 @@ If you would like to contribute a feature then fork the master branch (fork the
 ```
 nosetests -s -v tests/
 ```
-Let us know what you want to do just in case we're already working on an implementation of something similar. This way we can avoid any needless duplication of effort. Also, please don't forget to add tests for any new functions.
```
jkleckner: @jmschrei Is this whole file redundant with the index.rst file and should it be removed?

jmschrei: I think that is the case.

jmschrei:
I agree with the intent of this PR. However, I unabashedly use tabs when coding, and so even if we fix this issue here it will pop up again as I change the code. I am thus willing to accept that there will be this noise in future PRs.

jkleckner:
@jmschrei,  I was careful not to remove any tabs that were not at the end of a line and did not expand any tabs.  It does seem that many of the contributors are deleting trailing white space and this would be a non-semantic change set to eliminate those extraneous lines of change in the future such as those in #277

But to add a little humor check out this blog: https://stackoverflow.blog/2017/06/15/developers-use-spaces-make-money-use-tabs/

docs/index.md~
jkleckner: Ok, I added a removal of it to the patch set.

jkleckner:
Out of date now anyway.

PR #277, opened 2017-06-16
gmm + hmm: Add Callbacks and History

mrshu:
* Add Callbacks class that provides the ability for code to be executed
  on the start/end of training and start/end of each training iteration.

* Add a special callback History that holds historical information on
  training (time per iteration, improvement per iteration, ...).

* Since this PR aims to also include timing info in the History objects, it contains also #269. Upon merging this is, #175 should be resolved.


pomegranate/gmm.pyx
```diff
@@ -138,48 +143,99 @@ cdef class GeneralMixtureModel(BayesModel):
...
+		self.stop_training = False
 
 		initial_log_probability_sum = NEGINF
 		iteration, improvement = 0, INF
 
-		while improvement > stop_threshold and iteration < max_iterations + 1:
+		training_start_time = time.time()
+		while True:
+			callbacks.on_iteration_begin(iteration)
+			if self.stop_training:
```
jmschrei: Shouldn't this happen at the end of the loop, as the improvement is calculated at the end?

pomegranate/callbacks.py
```diff
@@ -0,0 +1,97 @@
...
+        self.history = {}
+        self.iterations = []
+
+    def on_iteration_end(self, iteration, logged_vars=None):
+        self.iterations.append(iteration)
+        for var, val in logged_vars.items():
+            self.history.setdefault(var, []).append(val)
+
+
+class ParamsTerminator(Callback):
```
jmschrei: The point of this is to replace the use of a while loop? I think that loop should remain and the callbacks are additional things that you'd like to do. Otherwise the termination conditions need to be included in every callback you write. Perhaps we could have a property from the callback `continue_training` that defaults to true, and if it ever defaults to False then it stops training, overriding the other conditions?

jmschrei:
Thanks for the great contributions. These are very useful. I think this PR is being merged with your other, unfortunately, so let's resolve that first.

mrshu:
Thank you for your kind comments @jmschrei -- I agree on the plan, let's first resolve #269

pomegranate/gmm.pyx
mrshu: @jmschrei I put it there in order for the new code to be as close to the previous implementation as possible. But I agree that the better place for this may be at the end of the loop.

pomegranate/callbacks.py
mrshu: > The point of this is to replace the use of a while loop?

To some extend, yes. The intention was for the `while` loop to stay there, but for all the stop conditions to be evaluated as part of the callbacks subsystem. This allows the stopping code to be reused better, and also the training code (the `while` loop) to become a bit cleaner.

> I think that loop should remain and the callbacks are additional things that you'd like to do.

Ok, I do not think I agree but I am happy to go with whatever you decide to do.

> Otherwise the termination conditions need to be included in every callback you write. 

I am sorry, I do not quite follow here.

I am not sure why do you think that the termination conditions need to be included in every callback one would write, but I agree that the `ParamsTerminator` would have to be included in the list of callbacks in order for the training to finish. The implementation in this PR takes care of that by adding the `ParamsTerminator` to the callback list in all cases.

> Perhaps we could have a property from the callback continue_training that defaults to true, and if it ever defaults to False then it stops training, overriding the other conditions?

Sorry, I do not quite follow here. Could you please expand on what you meant by this?

pomegranate/gmm.pyx
```diff
@@ -232,7 +288,7 @@ cdef class GeneralMixtureModel(BayesModel):
 		if not self.is_vl_:
 			X_ndarray = _check_input(X, self.keymap)
 			X_ptr = <double*> X_ndarray.data
-			
+
```
jkleckner: This and many lines of difference below are an example of the noise eliminated by #273.

jmschrei:
If you could rebase this on top of the current version I'll take a look again.

mrshu:
@jmschrei will do.

If you like the general approach of having these callbacks everywhere, we can add them to the same classes that have had the timing info added in #269

jmschrei:
I like the general approach of having the callbacks everywhere, and I like having the user be able to specify more complicated stop criterion, but I don't want that to supercede the max_iterations or min_iterations parameters. Essentially, if someone creates a callback to send them an email when an iteration is done, I don't want them to inadvertently now have their model train forever.

jmschrei:
Thanks for the first draft of this! It was very helpful for my attempt.  I incorporated and extended these ideas in the other PR.

PR #279, opened 2017-06-24
Correcting code documentation for NaiveBayes.

MGwynne:
When I was looking through the API documentation for NaiveBayes on ReadTheDocs, I noticed that:

  1. The pseudocount parameter documentation for from_samples was tabbed incorrectly (spaces
      rather than tabs), inconsistent with the rest of the docs, which broke the HTML version.
  2. There was an n_components parameter, which seems to have been removed and is now
      inferred automatically.
  3. The n_jobs parameter wasn't documented (but was in e.g., fit).

So I thought I'd fix the above issues (tabbed, removed + added) when I encountered them.


jmschrei:
Thanks for the contribution! It looks like I had just copied the `from_samples` method from the mixture models without actually updating everything.

PR #288, opened 2017-07-06
[MRG] Added option for explicit end states to hmm.from_samples()

nachtm:
Added the option end_state=False to hmm.from_samples. If true, the HMM will also train to determine which states are likely to be the end of the sequence.


jmschrei:
This LGTM but let's debug the issue you're facing before merging this.

jmschrei:
Would you mind adding in a few unit tests in tests/test_hmm.py? Just basically learning a hmm with explicit ends or not and checking to make sure that the learned structure includes or doesn't include those transitions?

jmschrei:
Great, thanks!

PR #296, opened 2017-07-19
remove old FiniteStateMachine examples

ivanov:
Back in  8a0d5f368a21a25c6b41b2697a95bc56be0d9b1c,  @jmschrei had removed the code for FiniteStateMachines, but all of these examples were still around.


jmschrei:
Great, thanks for taking care of it. I'll let the tests run just to be sure and then merge.

PR #301, opened 2017-08-02
Fix typo in distributions

kwmsmith:
Remove `weights=[...]` in example.

### DIFF
--- a/docs/Distributions.rst
+++ b/docs/Distributions.rst
@@ -64,7 +64,7 @@ However, frequently we don't know the parameters of the distribution beforehand 
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


jmschrei:
Thanks for the submission! Why is this a typo?

kwmsmith:
I'm inferring from the example and text that follows:

> If we want to fit the model to weighted samples, we can just pass in an array of the relative weights of each sample as well...

which only makes sense if `weights` isn't included in the first `NormalDistribution` example, at least to my reading...

jmschrei:
You're totally right, thanks. I was looking at the document after your modifications and thinking "this looks right though..."

Issue #314, opened 2017-08-17
Compilation error on Ubuntu 16.04: missing base.c


ucsky:
When I try to:

`python3 setup.py install`

I get the following error message:  

`x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-
  strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -
  I/usr/include/python3.5m -I/usr/local/lib/python3.5/dist-packages/numpy/core/include -c 
  pomegranate/base.c -o build/temp.linux-x86_64-3.5/pomegranate/base.o
  x86_64-linux-gnu-gcc: error: pomegranate/base.c: No such file or directory
  x86_64-linux-gnu-gcc: fatal error: no input fil`

It's look like base.c is missing.

Any idea why?

tuxdna:
I observe the same behavior on Fedora release 25.

```
$ python setup.py clean build
running clean
removing 'build/temp.linux-x86_64-3.5' (and everything under it)
running build
running build_py
running egg_info
writing pomegranate.egg-info/PKG-INFO
writing requirements to pomegranate.egg-info/requires.txt
writing dependency_links to pomegranate.egg-info/dependency_links.txt
writing top-level names to pomegranate.egg-info/top_level.txt
reading manifest file 'pomegranate.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'pomegranate/*.c'
writing manifest file 'pomegranate.egg-info/SOURCES.txt'
running build_ext
building 'pomegranate.base' extension
creating build/temp.linux-x86_64-3.5
creating build/temp.linux-x86_64-3.5/pomegranate
[...]

Issue #313, opened 2017-08-16
import pomegranate from python3 fail with Ubuntu 16.04


ucsky:
After installing pomegranate on Ubuntu 16.04 with
  sudo pip3 install pomegranate  

I get the following error message when I try to import pomegranate
>>> import pomegranate
  Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/pomegranate/__init__.py", line 9, in <module>
  from .base import *
  File "pomegranate/base.pyx", line 1, in init pomegranate.base (pomegranate/base.c:11221)
  ImportError: No module named 'pomegranate.utils'

Please any help with this?

jmschrei:
This can happen when you have an older version of pomegranate already installed and messes with the install. Can you try deleting pomegranate completely from your computer (including manually from site-packages) and reinstalling it?

ucsky:
I did:

`sudo pip3 uninstall pomegranate`
`sudo pip3 install pomegranate --no-cache-dir`

but I still have the same error message.

fpdx:
I'm having the same problem with `pip3`. I cleaned everything by hand, but the problem is still there.

I see [114](https://github.com/jmschrei/pomegranate/issues/114) and [215](https://github.com/jmschrei/pomegranate/issues/215).
numpy could be the problem (but I cannot downgrade it now as suggested in 215): I have `numpy 1.13.1`, and `1.12.0` was already problematic. Is there a chance to solve this? Thank you.

ucsky:
I installed on different computer with Ubuntu 16.04. On some computer pomegranate look to work fine but on other not. I checked the numpy version it's all 1.13.1

jmschrei:
I don't immediately know what the issue could be. Are you -positive- that you've removed all remnants of pomegranate before upgrading? There might be an egg file or something hanging around. Can you also try cloning from github and letting me know if that works?

fpdx:
I uninstalled both with `pip uninstall pomegranate` and `pip3 uninstall pomegranate`. Then removed by hand all the output files of `sudo find / -name pomegra*`, until it gives no output. Then run `sudo pip3 install pomegranate --no-cache-dir`. Then in `ipython3` the `import pomegranate` gives the error:

    ---------------------------------------------------------------------------
    ImportError                               Traceback (most recent call last)
    <ipython-input-1-f115d95d8f94> in <module>()
    ----> 1 import pomegranate
[...]

PR #317, opened 2017-08-31
Update README for installation instruction on Fedora

tuxdna:


tuxdna:
Some background can be found here:

 * https://github.com/jmschrei/pomegranate/issues/314
 * https://github.com/jmschrei/pomegranate/issues/313

README.md
```diff
@@ -38,6 +38,27 @@ To run the tests, you also must have `nose` installed.
 
 ### User Installation
 
+#### Fedora Release 25 onwards
```
jmschrei: I think this should be moved to after the other OS's

README.md
```diff
@@ -38,6 +38,27 @@ To run the tests, you also must have `nose` installed.
...
+
+Install Pomegranate and its dependencies ( for Python3 )
+
+```
+$ sudo dnf install -y python3-virtualenv python3-scipy python3-numpy python3-networkx python3-Cython
+$ virtualenv env3 --python=python3.5 --system-site-packages
```
jmschrei: Is it possible to do this without a virtualenv? It seems like it might throw off some newer users by adding the additional complexity of virtualenvs to the installation process.

README.md
```diff
@@ -38,6 +38,27 @@ To run the tests, you also must have `nose` installed.
...
+$ source env3/bin/activate
+(env3) $ pip3 install wheel nose cython numpy scipy networkx
+(env3) $ pip3 install --no-cache-dir pomegranate
+```
+
+Our virtualenv is ready, so now we can check that the installation is ready
+
+```
+(env3) $ python -c 'from pomegranate import *; print("It works")'
+It works
```
jmschrei: Maybe this should be changed to something more like `python -c 'from pomegranate import *; print(NormalDistribution(0, 1).probability(0))`

README.md
```diff
@@ -38,6 +38,27 @@ To run the tests, you also must have `nose` installed.
...
+```
+
+#### Other OSes
```
jmschrei: If the `Fedora Release 25 Onward` is moved after this, I don't think this line is needed.

jmschrei:
This is great, thanks for the addition! If you can make the suggested changes I'd love to merge this.

tuxdna:
@jmschrei I have updated the README file as per your review comments. I hope it is good to merge this time :-)

jmschrei:
Perfect! Thanks for the additions, this is very helpful.

Issue #320, opened 2017-09-06
BayesianNetwork.from_structure() does not give the expected name to the network


handsaway:
EDIT: the output I posted before was not corresponding to the input, I updated with the right output, where the wrong behavior is visible.

When creating a model using BayesianNetwork.from_structure(), the resulting network does not have the name given in argument, but gets the name of the last node instead.

Example:
```
from pomegranate import *

data = [['A', 'A', 'A'],
        ['A', 'A', 'A'],
        ['A', 'A', 'A'],
        ['A', 'A', 'A'],
        ['A', 'A', 'A'],
        ['B', 'B', 'B'],
        ['B', 'B', 'C'],
        ['C', 'C', 'A'],
        ['C', 'C', 'C'],
        ['C', 'C', 'C'],
        ['C', 'C', 'C'],
        ['C', 'B', 'A']]

structure1 = ((), (0,), (0, 1))
network = BayesianNetwork.from_structure(data, structure=structure1, name='monty_hall_network', state_names=['guest', 'prize', 'host'])
network.fit(data)

print(network.to_json())
```
gives a BayesianNetwork with name **'host'** (name of the last node) instead of **'monty_hall_network'**.
I have checked in debug that this behavior is not due to to_json() since the wrong name is attributed before using to_json().

Output:
```
{
    "class" : "BayesianNetwork",
    "name" : "host",
    "structure" : [
        [],
        [],
        [
            0,
            1
        ]
    ],
    "states" : [
        {
            "class" : "State",
            "distribution" : {
                "class" : "Distribution",
                "name" : "DiscreteDistribution",
                "parameters" : [
                    {
                        "A" : 0.4166666666666667,
                        "B" : 0.16666666666666666,
                        "C" : 0.4166666666666667
                    }
                ],
                "frozen" : false
            },
            "name" : "guest",
            "weight" : 1.0
        },
        {
            "class" : "State",
            "distribution" : {
                "class" : "Distribution",
                "name" : "DiscreteDistribution",
                "parameters" : [
                    {
                        "A" : 0.4166666666666667,
                        "B" : 0.25,
                        "C" : 0.3333333333333333
                    }
                ],
                "frozen" : false
            },
            "name" : "prize",
            "weight" : 1.0
        },
        {
            "class" : "State",
            "distribution" : {
                "class" : "Distribution",
                "name" : "ConditionalProbabilityTable",
                "table" : [
                    [
                        "A",
                        "A",
                        "A",
                        "1.0"
[...]

PR #326, opened 2017-09-18
Rename variable to avoid overriding function argument

tuxdna:
Fixes: https://github.com/jmschrei/pomegranate/issues/320

More details are here: https://github.com/jmschrei/pomegranate/issues/320#issuecomment-330333122


jmschrei:
Thanks!

PR #328, opened 2017-09-22
Specify networkx version <2.0.

nlzimmerman:
Networkx 2.0 changed the behavior of graph.nodes() to produce a NodeView
(which cannot be numerically indexed) instead of a list. See, e.g.
https://networkx.github.io/documentation/networkx-1.10/_modules/networkx/classes/graph.html
https://networkx.github.io/documentation/networkx-2.0/_modules/networkx/classes/graph.html

It's a bummer.

The simplest fix is to require a pre-2.0 version of networkx, which is
all this patch does.

The specific place where this crashes is in hmm.pyx line 734 and 735:
```python
            prestates = self.graph.nodes()
            indices = { prestates[i]: i for i in range(len(prestates)) }
```
doesn't work because prestates[i] doesn't work, because prestates isn't a list under networkx 2.0.

I have not researched how much work it would be to rewrite this such that it's networkx 2.0 compatible. Sorry.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 numpy >= 1.8.0
 joblib >= 0.9.0b4
-networkx >= 1.8.1
+networkx >= 1.8.1, < 2.0
 scipy >= 0.17.0


jkleckner:
FYI, you can test the version in the following way:
```
import networkx as nx
print(nx.__version__)
old_nx = nx.__version__.startswith("1.")

G = nx.complete_graph(5)

if old_nx:
    print(G.nodes())
else:
    print(G.nodes)

# Using the expression form:
n = (G.nodes if old_nx else G.nodes())
print(n)
```
Which worked with 1.11 and 2.0.

jmschrei:
There are other issues I am trying to resolve with networkx, such as how to get topological sort working in the same way as before... Thanks for the contribution.

PR #344, opened 2017-10-30
Fixed tabbing issue in pseudocount documentation

andim:
The docstrings in gmm.pyx use spaces instead of tabs for the pseudocounts description which breaks the rendering of the documentation (see e.g. http://pomegranate.readthedocs.io/en/latest/GeneralMixtureModel.html#pomegranate.gmm.GeneralMixtureModel.fit). This PR fixes this issue.


jmschrei:
Great, thanks for the fix.

PR #359, opened 2017-11-26
Fix TypeError in HiddenMarkovModel.from_samples()

fortiema:
When using `HiddenMarkovModel.from_samples()` with `DiscreteDistribution` and passing custom labels, if name attribute is left to None a TypeError will be raised from the concatenation with string.

- Cast `name` to str to prevent "NoneType + str" TypeError when name is None
- Add test case to prevent regression

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -17,7 +17,8 @@ Changelog
 HiddenMarkovModel
 -----------------
 
-	- The documentation has been fixed so that states are defined as `State(NormalDistribution(0, 1))` instead of incorrectly as `State(Distribution(NormalDistribution(0, 1)))
+	- The documentation has been fixed so that states are defined as `State(NormalDistribution(0, 1))` instead of incorrectly as `State(Distribution(NormalDistribution(0, 1)))`
+	- Fixed a bug in `from_samples` that was causing a TypeError if `name` was not specified when using `DiscreteDistribution` with custom labels.
 
 
 Distributions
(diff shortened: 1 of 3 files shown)


fortiema:
Any idea why the appveyor build for python 3.4 fails?

jmschrei:
I haven't seen that error before. Quickly googling it suggests that there might've been an issue in Cython 0.27.0 that was fixed in 0.27.1/0.27.2, but we're using 0.27.3... I tried rebuilding the PR, let's see if it was just an anomaly.

fortiema:
The plot thickens... Seriously I don't have much experience with appveyor and Win builds. Seems to be an issue with the scipy build being used from a quick look at google results, but again I wouldn't know much about Win environments.

jmschrei:
Can you try pinning the requirement to Cython=0.25.0 and see if it runs? You'll need to update the dev-requirements.txt and the setup.py file.

fortiema:
Seems like it's not making any difference.

fortiema:
Reverted last commit. I've seen the other open PR also fails the 3.4 appveyor build as well.

jmschrei:
Yeah this doesn't seem like an issue on your part. Gonna go ahead and merge, thanks for the contribution.

jmschrei:
Actually, would you mind updating the docs/whats_new.rst file with this quick fix? It should be under the 0.8.2 changes on the top.

fortiema:
Sure thing, just did. BTW I noticed some dependencies are missing to be able to build the docs (sphinx, sphinx_rtd_theme, mock). Would probably be better to add it to dev-requirements.txt. I can submit a PR for that as well later.

jmschrei:
Great! I'll merge once the other tests pass.

jmschrei:
Thanks for the contribution!

PR #364, opened 2017-12-03
Fix dev requirements for docs

fortiema:
* Add sphinx, sphinx-rtd-theme to dev-requirements.txt
* Comment out the `mock` import in `docs/conf.py` as it is not used by default and mock is not in the requirements file
* Add `.python-version` to gitignore


jmschrei:
Thanks!

PR #406, opened 2018-03-28
Add zip_safe=False to setup.py

jkleckner:
If someone installs pomegranate into site-libraries using the
traditional command of "python setup.py install", the sources will
install as a zipped egg without this option.  Current cython will not
work with cimport from the zipped file.

With this change, users will be protected from seeing this error when
installing in this way.


jmschrei:
Great, thanks!

PR #424, opened 2018-04-30
Add Makefile for convenience in development

jkleckner:
This Makefile will invoke build, install, test, and uninstall on
conda environments for python 2.7 and 3.6.

An nbtest target to use the "jupyter nbconvert --execute" functionality
to find any errors in example and tutorial notebooks.

A minor change to tutorials/Tutorial_3_Hidden_Markov_Models.ipynb
sets the kernelspec to be python2 similar to the other notebooks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,87 @@
+PACKAGE_NAME=pomegranate
+PY2_ENV=py2.7
+PY3_ENV=py3.6
+
+default:
+	echo no default
+
+.PHONY: install test bigclean nbtest nbclean
+.PHONY: bigbuild py2build py3build
+.PHONY: biginstall py2install py3install
+.PHONY: biguninstall py2uninstall py3uninstall
+.PHONY: bigtest py2test py3test
+.PHONY: bignbtest py2nbtest py3nbtest
+
+install:
+	python setup.py install
+
+test:
+	python setup.py test
+
+bigclean: nbclean
+	rm -rf build
+	rm -rf dist
+	rm -rf .eggs
+	rm -rf $(PACKAGE_NAME).egg-info
+	find . -name '*.pyc' -print | xargs rm
+	find . -name '*.so' -print | xargs rm
+ifndef NO_REMOVE_CFILES
+	find . -name '*.c' -print | xargs rm
+endif
+
+# Add python dependencies
+ifdef PY2_ENV
+bigbuild: py2build
+biginstall: py2install
+biguninstall: py2uninstall
+bigtest: py2test
+bignbtest: py2test
+endif
+
+ifdef PY3_ENV
+bigbuild: py3build
+biginstall: py3install
+biguninstall: py3uninstall
+bigtest: py3test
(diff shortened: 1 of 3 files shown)


jmschrei:
Awesome. Thanks for your continued contributions. What is your suggested usage of this?

jmschrei:
Can you make a quick update to docs/whats_new.rst, adding in a Tutorials section and a Makefile section, describing the changes briefly?

jkleckner:
I find that I have two conda environments, one for legacy 2.7 python and one for current stable 3.6 python.  This makes it quick to grab a new pomegranate release and build/test/replace-install for those two environments.  You can move around on branches and deploy that version into both (or just one) environments.

I noticed people finding errors in the examples and tutorials and it occurred to me that it should be possible to run those as part of the testing process.  So I adapted my personal Makefile to include an `nbtest` target to test those notebooks.  At that point, I thought it might be useful to you or others to perform a sanity check over those assets periodically.  If those notebooks stabilize, in principle the Travis build could include them in the release testing as well.

I can write something along these lines since you are receptive to including it (I didn't want to presume).

[Here is a gist](https://gist.github.com/jkleckner/bb116786eb692937cc015030a76eff5f) created with `make py27nbtest` that stops on first error for each notebook.  If you `--allow-errors`, then all of the cells will be evaluated and you can manually view the html renderings to see the error messages for failing code.

Eventually, it might be possible to upgrade the notebooks to work with either python 2 or python 3 at which point modifying the kernelspec would allow them to be evaluated both ways.

jkleckner:
@jmschrei Ok, I replaced the patch with on rebased to master so that the whats new information would be in the right place.

jkleckner:
Completely correct.  I use PyCharm for development in Python and IntelliJ for Scala (and couldn't live without).  There is nothing sacred about the environment names and, in retrospect, I suppose they could be just called "python2" and "python3" for the two major environments.  I only use this Makefile when fully cleaning/testing/reinstalling into the two environments, most often right after a git branch change.  Let me know if you want to change the names to blur the specific version choices.  It is pretty quick to completely re-create a conda environment since all the packages are cached.

Makefile
```diff
@@ -0,0 +1,87 @@
+PACKAGE_NAME=pomegranate
+PY27_ENV=py2.7 # The conda env set up for python 2.7
```
jkleckner: I forgot that you can't have inline comments with make.  In trying to comment this for others, I introduced a bad bug that causes the env to include a space.  I will replace this patch to get rid of the bad version.

jkleckner:
@jmschrei  Note that if you run `make nbtest` on master that it generates errors such as the following:
```
pomegranate/base.pyx in init pomegranate.base
ImportError: No module named distributions
```

jkleckner:
See #426 for fix to setup.py

jmschrei:
Thanks for the thorough description! Let's give it a whirl. I'm fine with explicit names as long as it's documented somewhere.

PR #428, opened 2018-05-01
Ipynb fixes

jkleckner:
Note that it may be a regression that the list of states/nodes can't be passed in.  If that is fixed, then those changes here aren't needed.


jkleckner:
A nice goal for the next release would be to achieve a clean `make nbtest`

jmschrei:
Yes it would. I've been meaning for a while to rewrite a lot of the tutorials to make them cleaner and focus on more recent features. This will be a good guide to help. Thanks for your contributions!

jkleckner:
My apologies for the churn in this patch set. I'm used to a rebase rather than merge style.  I will rebase this to master then stop fiddling...

jmschrei:
I'm not sure what's up with Travis, but it seems to be disconnecting a lot on one build in particular...

PR #439, opened 2018-05-24
Convert from random.random to numpy random_sample

jkleckner:
- Replaced the uses of random.random() with numpy.random.random_sample()
  so that numpy is consistently used throughout the implementation of
  pomegranate.
- One concern was raised about the speed enhancement of viterbi which
  depends on setting the random seed to a deterministic value.  Doing
  this could affect the quality of random number generation during
  non-testing use of pomegranate.  This was introduced in PR#118
- Note this discussion about the thread safety of numpy.  When joblib is
  used for parallelization and a seed has been set, then the jobs will
  not generate diversity from each other in their samples.
  - https://stackoverflow.com/a/31058798

### DIFF
--- a/pomegranate/distributions/ConditionalProbabilityTable.pyx
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pyx
@@ -17,7 +17,6 @@ from ..utils import check_random_state
 import itertools as it
 import json
 import numpy
-import random
 import scipy
 
 from collections import OrderedDict
--- a/pomegranate/distributions/DiscreteDistribution.pyx
+++ b/pomegranate/distributions/DiscreteDistribution.pyx
@@ -8,7 +8,6 @@ import numpy
 import sys
 import itertools as it
 import json
-import random
 
 from libc.stdlib cimport calloc
 from libc.stdlib cimport free
--- a/pomegranate/distributions/GammaDistribution.pyx
+++ b/pomegranate/distributions/GammaDistribution.pyx
@@ -6,7 +6,6 @@
 
 import numpy
 import scipy
-import random
 
 
 from ..utils cimport _log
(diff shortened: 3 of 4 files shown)


jkleckner:
I will not be offended if you reject this patch set.

It just makes the use of random number generator consistent inside of the implementation of pomegranate.

I am a bit concerned about the setting of the seed to zero as a side-effect of Viterbi training:
- https://github.com/jmschrei/pomegranate/pull/439/files#diff-16503312cc32dfb87898df1d62fd1d8bR881
Won't that affect the randomness of all subsequent calls?

jmschrei:
I support converting everything over to numpy. The right thing to do is allow users to pass in their own seed, but that might be a bigger PR.

As for affecting randomness, yes it does. If I recall, the reason that I did that was because you'd get a different ordering of states when doing toplogical sort if you don't set the seed, which I found troublesome and potentially confusing for users. What are your thoughts?

jkleckner:
Can you think of a different way of ordering the states in a canonical way?

I do understand the tension between quality randomness and user confusion and I now understand why my states were coming out different every time I ran a fit so that my test output could not be compared.  I certainly wanted determinism.  Because I didn't use Viterbi it didn't become consistent.  

I was thinking that the JSON output might usefully be sorted instead of left to hash order for similar reasons of testing comparisons, but that is a different topic.

That [SO discussion](https://stackoverflow.com/a/31058798) talks about how the API of sklearn includes kwargs arguments for passing in the RandomState which is probably also desirable.

I would kind of lean towards not setting seeds in the library itself to prevent surprises caused by lack of diversity among jobs.

jkleckner:
See related discussion in #499

echo66:
If pomegranate uses stochastic processes inside functions to be distributed by joblib to different processes, I would strongly advise the PR to create, in advance, a list of integers and use each integer as a seed for each function execution. This will require a bigger PR.

jkleckner:
@echo66  I agree about `joblib` as mentioned in the first post in the thread.  This patch set was limited to just using one consistent source of entropy but it would be desirable to address to two competing goals of good entropy vs repeatability during testing.  Ideas?

PR #441, opened 2018-05-24
Add to_yaml convenience functions

jkleckner:
YAML is a superset of JSON that is 4 to 5 times more compact which makes
it easier to inspect visually.  To use it you have to make sure the yaml
package is installed.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -164,6 +164,8 @@ Base
 	
 	- Parameters `column_idx` and `d` have been added to the `_summarize` method that all models expose. This is only useful for univariate distributions and models that fit univariate distributions and can be ignored by other models. The `column_idx` parameter specifies which column in a data matrix the distribution should be fit to, essentially serving as an offset. `d` refers to the number of dimensions that the data matrix has. This means that a univariate distribution will fit to all samples `i` such that `i*d + column_idx` in a pointer array. Multivariate distributions and models using those can ignore this.
 
+	- A convenience function `to_yaml` was added to `State` and `Model` classes.  `YAML` is a superset of `JSON` that can be 4 to 5 times more compact.  You need the `yaml` package installed to use it.
+
 
 Distributions
 -------------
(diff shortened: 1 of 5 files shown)


jmschrei:
Thanks for the functions! This seems like a good thing to have. Is it possible to add a test for each model? You can probably copy the `to_json` and `from_json` unit tests for each model but use `to_yaml` and `from_yaml` instead.

jkleckner:
I intentionally didn't create a `from_yaml` because I wasn't sure that it would be as reliable as `from_json`.

Json serialization/deserialization is heavily relied on and I almost don't want to encourage people to use yaml for anything other than human readability.  I may be too cautious, though, since yaml is also widely used.

jmschrei:
How difficult would it be to add in a from_yaml method? One benefit of including many deserialization techniques is that it makes pomegranate integrate more with other packages or programs that may use that serialization/deserialization scheme

jkleckner:
The lazy way I layered it on top of `to_json` means that the same trivial thing could be done for `from_json`.  Specifically just read the `YAML`, convert to `JSON`, then call `from_json`.

I suppose tests could be done that way leveraging the existing `JSON` data as a reference.

jmschrei:
I think it would be helpful to have a `from_yaml` format as well and to include a test. I can also imagine a situation in which someone writes out a model, uses the yaml format due to its readability, tweaks it manually, and would like to read it back in.

jkleckner:
That failure was an example of the whitespace conflict resolved in #445 ...
Once rebased that should be a lot better...

jkleckner:
I force-pushed a branch that should have resolved the conflict on whitespace.  I don't understand why github is still showing a conflict.

jkleckner:
I resolved the wrong line - now testing.

jmschrei:
Looks good! I'm fine with the pyyaml requirement, it seems easy enough to install.

jmschrei:
Looks like this killed the autodocs. We needed to add `pyyaml` to the `dev-requirements.txt` file as well.

PR #442, opened 2018-05-24
Add hmm sample_state function

jkleckner:
This will sample a HiddenMarkovModel given a state vector
such as returned by predict_proba

### DIFF
--- a/pomegranate/hmm.pyx
+++ b/pomegranate/hmm.pyx
@@ -1298,6 +1298,119 @@ cdef class HiddenMarkovModel(GraphModel):
 
         return numpy.array(emissions)
 
+    def sample_state(self, state_probs, length=1):
+        """Generate samples from a model in a given emission state
+
+        Returns a list of feature vectors.
+        The model must have been baked first in order to run this method.
+
+        Parameters
+        ----------
+        state_probs : array-like, shape (n_states)
+            The normalized probabilities of being in each state such as those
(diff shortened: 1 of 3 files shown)


pomegranate/hmm.pyx
```diff
@@ -1279,6 +1279,119 @@ cdef class HiddenMarkovModel(GraphModel):
...
+        """Generate samples from a model in a given emission state
+
+        Returns a list of feature vectors.
+        The model must have been baked first in order to run this method.
+
+        Parameters
+        ----------
+        state_probs : array-like, shape (n_states)
+            The normalized probabilities of being in each state such as those
+            returned from predict_proba
```
jkleckner: @jmschrei  Note that predict_proba returns a subset of states (the visible ones).  Do you think a `sample_state` should do that as well?

jmschrei:
Oh, good idea. Do you think this should be part of `sample` where you can pass in, optionally, a list of states? Looking at the API, I should actually make this the `path` input, and change what `path` does currently to `return_path`

jkleckner:
@jmschrei Interesting idea to overload path.  But for some reason, it doesn't feel quite right to me.  One aspect of `sample_state` is that it can return numerous samples without advancing the state while `sample` advances the model through states, so they seem like different semantics.

I still wonder whether the semantics of `sample_state` should allow specifying a hidden state for the sampling...

PR #443, opened 2018-05-24
Change init in bayes and gmm to use sum(weights)

jkleckner:
In bayes.pyx and gmm.pyx, use sum(weights) instead of weights.sum() in
__init__() to permit lists to be passed in.


jmschrei:
Sounds good to me. Are there other such instances where this could be updated?

jkleckner:
I did a "git grep" and all of the other uses were properly using np.array for the `sum()`

jmschrei:
Great! Thanks for the contribution.

PR #445, opened 2018-05-24
Remove whitespace at end of line

jkleckner:
This is a non-semantic patch set that does not change tabs but cleans
out whitespace from the end of lines.

The motivation for this is that PyCharm does this cleanup when editing
files and it results in extra noise when creating and managing patch
sets because you need to constantly be doing a "git diff -b" to filter
away the extraneous noise.

To verify this patch set changes nothing semantic you can run
"git diff -b HEAD^" and observe no output.


jkleckner:
No leading tabs were hurt during the making of this patch set ;-)

jmschrei:
+1

PR #447, opened 2018-05-24
Add test for hmm using gmm from_json

jkleckner:
This patch set allows GeneralMixtureModels to be used with
HiddenMarkovModel.from_json() and adds a test.

It is a bit of a mystery why the GeneralMixtureModel class is not in
scope already.

### DIFF
--- a/pomegranate/NaiveBayes.pyx
+++ b/pomegranate/NaiveBayes.pyx
@@ -83,7 +83,7 @@ cdef class NaiveBayes(BayesModel):
 		nb = {
 			'class' : 'NaiveBayes',
 			'models' : [json.loads(model.to_json()) for model in self.distributions],
-			'weights' : self.weights.tolist()
+			'weights' : numpy.exp(self.weights).tolist()
 		}
 
 		return json.dumps(nb, separators=separators, indent=indent)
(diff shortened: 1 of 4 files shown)


jkleckner:
@jmschrei would you please look into the test failure to see if something obvious in the process is wrong?

jmschrei:
Nice, thanks for the contribution. I'm guessing the issue is that the weights are stored as the log weight instead of the actual value. Let me see.

jmschrei:
I think this needs to be changed to `numpy.exp(self.weights).to_list()`: https://github.com/jmschrei/pomegranate/blob/master/pomegranate/gmm.pyx#L424

jmschrei:
Same here, actually: https://github.com/jmschrei/pomegranate/blob/master/pomegranate/NaiveBayes.pyx#L86

jmschrei:
I think the right thing to do is for me to make a self.weights and a self.log_weights and use them appropriately-- but this will be a quick fix that will solve people's issues.

jkleckner:
@jmschrei Then you might want to take over this patch set?

jkleckner:
@jmschrei I amended the patch with those two fixes and the test now passes.

jmschrei:
Awesome, thanks!

PR #450, opened 2018-06-08
Memory leak in BayesModel._log_probability

hyeshik:
Hello,

Thank you for the great software. Recently, I found my code using pomegranate consume a lot of memory, usually more than 100 GiB, after thousands of calls to `hmm.viterbi()`. This change fixes the problem, and its memory footprint stays at a pretty stable level. Can you reflect this patch in your future release?


jkleckner:
Good catch.  This prompted me to look for more as with #451

jkleckner:
@jmschrei Note that #451 doesn't duplicate this one so it would also make sense to include.

PR #457, opened 2018-06-22
Fix sampling of Gamma Distributions

ptigwe:
At the moment, sampling from the gamma distribution simply ignores the parameter
`n` and always returns a single sample from the distribution


jmschrei:
Good catch, thanks!

Issue #432, opened 2018-05-03
Questions regarding the Documentation


yh21equn:
Hello.

I have some questions regarding the documentation of pomegranate:
- Is it possible to create a multidimensional normal distribution or should the multivariate Gaussian distribution be used instead?  
- Is the min_iterations parameter in `HiddenMarkovModel.fit()` only relevant for the Baum-Welch-Algorithm or also for the Viterbi-Algorithm?
- How should the json output from `HiddenMarkovModel.to_json()` be interpreted?  I have especially difficulties with the parameters of the normal distribution (Which is the mean and which is the variance?) and the transition probabilities between the edges (Which are the edge indices and which are the weights?). I have attached an example for a json output I got from my model.
[json_example.txt](https://github.com/jmschrei/pomegranate/files/1969935/json_example.txt)

Thank you for your help!

jmschrei:
Howdy

1. You can make any model multivariate either by using a MultivariateGaussianDistribution or an IndependentComponentsDistribution. The ICD will model each feature independently using any type of probability distribution (doesn't have to be Normal, but it can, and it doesn't have to be homogenous, you can have many different types.)

2. min_iterations is relevant to both. It's not relevant for labeled training

3. I'm not sure exactly what you mean by this. They are the parameters of the learned distributions / transition matrix,

yh21equn:
Hello.

Thank you for your answer.

- To the second point: Is min_iterations (and max_iterations) relevant if I use labeled viterbi training?
- To the third point: I have figured out how to read the json output by now,

jkleckner:
@yh21equn I also found that using arrays for parameters rather than name/value pairs made it opaque.  That was part of my motivation for PR #441 because `YAML` at least compresses the information enough that as a human you can see the structure.  The meaning gets lost in the sea of `JSON`.  See, for example, your output attached below as `YAML`.
[json_example.yaml.txt](https://github.com/jmschrei/pomegranate/files/2035393/json_example.yaml.txt)

@jmschrei Since this is the second occurrence of confusion about realizing `IndependentComponentsDistribution` is the generalized way to achieve the specific case of a vector of gaussians, a couple of pointers in the api doc, summary doc, and perhaps tutorials where `MultivariateGaussianDistribution` is used may be warranted?

Perhaps something like:

[...]

PR #460, opened 2018-06-26
Add correlation comments in multivariate docs

jkleckner:
- Add wording that explains to use IndependentComponentsDistribution
  when no correlation
- Add all distributions to API doc

See comment in #432


jmschrei:
👍

PR #480, opened 2018-08-07
[WIP] new student t-distribution

sdia:


jmschrei:
Howdy @sdia. This looks like a great start. I think that as @arose13 mentioned that the distribution should take in both the mean and standard deviation as well.

jmschrei:
@sdia I'm currently working to support custom python based distributions. Hopefully when I finish that it won't be necessary to have cython implementations of everything. Thanks again for your efforts.

arose13:
lol @jmschrei we all just keep disappointing you 😢  😞

jmschrei:
Not at all. I feel it's more my fault due to the code being in Cython and not as easy to extend as it should be as it's not as well known a language. I'm hoping to fix that with the next patch so that Python objects can be dropped in. That way contributions can be done in Python when no Cython alternative exists.

arose13:
@jmschrei For sure if that was done (and also is your intention) I'll make basically every single distribution there is 😋

PR #504, opened 2018-10-17
[MRG] Add in deterministic sampling

jmschrei:
This PR should allow all models to deterministically sample through passing in a random seed in a style similar to sklearn, i.e., by passing either an integer or a numpy.random.RandomState object through the `random_state` parameter of each `sample` method.

Example:

```python
>>> d = NormalDistribution(5, 2)
>>> d.sample(random_state=0)
8.528104691935328
```

### DIFF
--- a/pomegranate/hmm.pyx
+++ b/pomegranate/hmm.pyx
@@ -29,6 +29,7 @@ from .callbacks import History
 
 from .utils cimport _log
 from .utils cimport pair_lse
+from .utils import check_random_state
 
 from libc.stdlib cimport calloc
 from libc.stdlib cimport free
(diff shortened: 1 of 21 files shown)


pomegranate/hmm.pyx
```diff
@@ -29,6 +29,7 @@ from .callbacks import History
 
 from .utils cimport _log
 from .utils cimport pair_lse
+from .utils import check_random_state
```
jkleckner: Can't seem to comment on line 883 in github.
I would like to comment that at least a comment is in order for that line.
There are serious repercussions to setting the seed to 0 as a side-effect of bake.
 https://github.com/jmschrei/pomegranate/pull/439/files#diff-16503312cc32dfb87898df1d62fd1d8bR881

PR #506, opened 2018-10-24
Fix typos in `HiddenMarkovModel` doc string example 

paulbrodersen:
I think you forgot a few zeros when defining the distributions in the example. I fixed those typos and recomputed the log-likelihood based on the new values. Also changed the print statement to be python 3 compatible.


jmschrei:
Whoops. Thanks!

PR #521, opened 2018-11-22
[MRG] Add custom user distribution support via Python fallback

jmschrei:
This PR enables support for custom Python distributions to be used in place of the built-in distributions, hopefully ending the long-running nightmare of not having all distributions implemented. This is done by checking whether the passed in distribution inherits from the distribution class, and if not, falls back to using its python functions. This requires that users implement the following methods in their shiny new distribution class:

* **log_probability(X)**: This method takes in a vector/matrix of observations and returns a vector of log probabilities
* **summarize(X, weights)**: This method takes in a vector/matrix of observations, X, and a vector of associated weights and stores sufficient statistics to the model.
* **from_summaries(self, inertia)**: This method will use the stored sufficient statistics in order to update model weights.
* **clear_summaries(self)**: This method simply resets the stored sufficient statistics.

The only attribute that the object has to have is *self.d*, which stores the number of dimensions being modeled.


pomegranate/hmm.pyx
```diff
@@ -1398,9 +1412,13 @@ cdef class HiddenMarkovModel(GraphModel):
         cdef numpy.ndarray f_ndarray = numpy.zeros((n+1, m), dtype=numpy.float64)
         cdef double* f
 
+        print(self.cython)
```
jkleckner: This and line 1420 look like debug scaffolding left in.

PR #557, opened 2019-02-22
Correct spelling of is_gpu_enabled in documentation

alexhenrie:
The version of this function with a leading underscore is only available to Cython code.


jmschrei:
Thanks!

PR #559, opened 2019-03-01
Minor Kmeans update

rchliu:
This change replaces an unnecessary list allocation and copy with an in-place truncation in the Kmeans code. It also updates the documentation for centroid initialization to account for the `kmeans||` initialization method introduced in 7bbad81.


jmschrei:
Seems reasonable.

Issue #589, opened 2019-06-30
Slow BNSL with missing data


Wesserg:
Hi. We are observing extreme slow down when running BNSL with missing variables (7days and still running). For the same dataset with imputed values the BNSL finishes within few hours. In both cases we are using 'exact' algorithm.

I was wondering if this is expected or is it a bug? Readthedocs mentions that the slowdown might occur in case of models based on multivariate normal distributions, not necessarily for Bayes nets. 

Could the issue be related to: [https://github.com/jmschrei/pomegranate/pull/571](https://github.com/jmschrei/pomegranate/pull/571) [We would appreciate a merge/pull if possible]

jmschrei:
I can't think of a reason why you would get a slowdown when running BNSL with missing variables. The only difference is that, when getting frequencies for each entry, you ignore those with missing values. Can you post a minimally reproducing example?

alexhenrie:
Here is a simple program that demonstrates the problem:

```
import numpy as np
from neurtu import delayed, Benchmark
from pomegranate import BayesianNetwork
from random import random

X = np.round(np.random.rand(10000, 15))
train = delayed(BayesianNetwork)().from_samples(X)

print('Without missing values:')
print(Benchmark(wall_time=True, cpu_time=True, repeat=3)(train))
print()

for i in range(X.shape[0]):
    for j in range(X.shape[1]):
        if random() <= 0.1:
            X[i][j] = np.nan

print('With missing values:')
print(Benchmark(wall_time=True, cpu_time=True, repeat=3)(train))
print()
```

With the current Pomegranate code, training a Bayes net with 10% of the values missing is nearly twice as slow as training a Bayes net without missing values:

```
Without missing values:
6it [00:23,  4.16s/it]
      cpu_time  wall_time
mean  4.655360   4.824831
max   4.715996   4.914933
std   0.054074   0.092098

With missing values:
6it [00:43,  7.60s/it]
      cpu_time  wall_time
mean  8.568754   8.596677
max   8.596781   8.667951
std   0.035506   0.090385
```

With the improvements I proposed in #571, the performance gap narrows slightly but is still almost as bad:

```
Without missing values:
6it [00:23,  4.19s/it]
      cpu_time  wall_time
mean  4.708834   4.766239
max   4.733736   4.840934
std   0.039952   0.093201

With missing values:
6it [00:42,  7.45s/it]
      cpu_time  wall_time
mean  8.439062   8.471104
max   8.558999   8.552451
[...]

PR #602, opened 2019-07-29
Add shortcuts for missing data

alexhenrie:
The following benchmarking program shows that on sklearn's ["digits" dataset](https://scikit-learn.org/stable/auto_examples/datasets/plot_digits_last_image.html), which is their dataset most suited for Bayesian analysis, adding these shortcuts reduces the total execution time by about 10% when there are missing values and does not increase the execution time when there are no missing values.

Interestingly, I could not reproduce the missing values slowdown on this dataset that I see with our own dataset or a randomly generated dataset. Nonetheless, the benchmarks show that there is a significant benefit to adding shortcuts.

Script:

```
import numpy as np
from random import random

train = delayed(BayesianNetwork)().from_samples(X)

print()

for i in range(X.shape[0]):
    for j in range(X.shape[1]):
        if random() <= 0.1:
            X[i][j] = np.nan

print('With missing values:')
print()
```

Without NaN shortcuts:

```
Without missing values:
      wall_time  cpu_time
mean  11.357033  7.907993
max   11.491907  7.944401
std    0.117067  0.031715

With missing values:
      wall_time  cpu_time
mean  10.816061  7.586988
max   10.831504  7.742489
std    0.013388  0.148912
```

With NaN shortcuts:

```
Without missing values:
      wall_time  cpu_time
mean  11.586937  7.937418
max   12.074349  8.004921
std    0.422112  0.067759

With missing values:
      wall_time  cpu_time
mean   9.708813  6.849663
max    9.726768  6.869913
std    0.020418  0.025905
```

Please let me know if there's anything else you need.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -2069,7 +2069,7 @@ cdef discrete_find_best_parents(numpy.ndarray X_ndarray,
 
 cdef double discrete_score_node(double* X, double* weights, int* m, int* parents,
 	int n, int d, int l, double pseudocount) nogil:
-	cdef int i, j, k, idx, is_na
+	cdef int i, j, k, idx
 	cdef double w_sum = 0
 	cdef double logp = 0
 	cdef double count, marginal_count
@@ -2080,23 +2080,21 @@ cdef double discrete_score_node(double* X, double* weights, int* m, int* parents
 	memset(marginal_counts, 0, m[d-1]*sizeof(double))
 
 	for i in range(n):
-		idx, is_na = 0, 0
+		idx = 0
 		for j in range(d-1):
 			k = parents[j]
 			if isnan(X[i*l + k]):
-				is_na = 1
-			else:
-				idx += <int> X[i*l+k] * m[j]
+				break
+			idx += <int> X[i*l+k] * m[j]
+		else:
+			k = parents[d-1]
 
-		k = parents[d-1]
+			if isnan(X[i*l+k]):
+				continue
 
-		if is_na == 1 or isnan(X[i*l+k]):
-			continue
-
-
-		marginal_counts[idx] += weights[i]
-		idx += <int> X[i*l+k] * m[d-1]
-		counts[idx] += weights[i]
+			marginal_counts[idx] += weights[i]
+			idx += <int> X[i*l+k] * m[d-1]
+			counts[idx] += weights[i]
 
 	for i in range(m[d]):
 		w_sum += counts[i]
--- a/pomegranate/distributions/ConditionalProbabilityTable.pyx
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pyx
@@ -183,18 +183,15 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 		return self.values[idx]
 
 	cdef void _log_probability(self, double* X, double* log_probability, int n) nogil:
-		cdef int i, j, idx, is_na = 0
+		cdef int i, j, idx
 
 		for i in range(n):
-			idx, is_na = 0, 0
+			idx = 0
 			for j in range(self.m+1):
 				if isnan(X[self.m-j]):
-					is_na = 1
-
+					log_probability[i] = 0.
+					break
 				idx += self.idxs[j] * <int> X[self.m-j]
-
-			if is_na == 1:
-				log_probability[i] = 0.
 			else:
 				log_probability[i] = self.values[idx]
 
--- a/pomegranate/distributions/ConditionalProbabilityTable.pyx
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pyx
@@ -277,7 +277,7 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 
 	cdef double _summarize(self, double* items, double* weights, int n,
 		int column_idx, int d) nogil:
-		cdef int i, j, idx, k, is_na
+		cdef int i, j, idx, k
 		cdef double* counts = <double*> calloc(self.n, sizeof(double))
 		cdef double* marginal_counts = <double*> calloc(self.n / self.k, sizeof(double))
 
@@ -285,26 +285,21 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 		memset(marginal_counts, 0, self.n / self.k * sizeof(double))
 
 		for i in range(n):
-			idx, is_na = 0, 0
+			idx = 0
 			for j in range(self.m+1):
 				k = i*self.n_columns + self.column_idxs_ptr[self.m-j]
 				if isnan(items[k]):
-					is_na = 1
-					continue
-
+					break
 				idx += self.idxs[j] * <int> items[k]
+			else:
+				counts[idx] += weights[i]
 
-			if is_na:
-				continue
+				idx = 0
+				for j in range(self.m):
+					k = i*self.n_columns + self.column_idxs_ptr[self.m-1-j]
+					idx += self.marginal_idxs[j] * <int> items[k]
 
-			counts[idx] += weights[i]
-
-			idx = 0
-			for j in range(self.m):
-				k = i*self.n_columns + self.column_idxs_ptr[self.m-1-j]
-				idx += self.marginal_idxs[j] * <int> items[k]
-
-			marginal_counts[idx] += weights[i]
+				marginal_counts[idx] += weights[i]
 
 		with gil:
 			for i in range(self.n / self.k):
--- a/pomegranate/distributions/ConditionalProbabilityTable.pyx
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pyx
@@ -259,25 +259,21 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 		self.__summarize(items, weights)
 
 	cdef void __summarize(self, items, double [:] weights):
-		cdef int i, n = len(items), is_na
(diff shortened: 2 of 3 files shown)


jmschrei:
This looks good. I'll review it more thoroughly when I get back next week from the conference I'm at. Would it be possible for you to confirm that you're getting the same network before and after your changes by printing out `model.structure` and by calculating the probability of some data set? It can be the one that learned the network over. Thanks!

alexhenrie:
I have now checked that the shortcuts result in no changes to model.structure or the output of model.predict_proba. I look forward to these commits being merged when you return from your conference, and after that I'll likely start digging into other areas where the performance could be improved. Thanks again for your support!

jmschrei:
When I run the test script on my machine I don't see the same speedups that you see.

This is the result before incorporating your change to `discrete_score_node` (which is generally the most compute intense part of structure learning).

```
Without missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.683641   8.551788
max   5.730023   8.625430
std   0.068112   0.064195

With missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.138517   7.745014
max   5.150633   7.810688
std   0.016490   0.057186
```

When I incorporate the change:

``
Without missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.744121   8.610118
max   5.888380   8.690713
std   0.137677   0.070891

With missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.517163   8.523160
max   5.574747   8.530496
std   0.066018   0.012541
```

To me, the differences look like normal run-to-run differences. 

When I cranked up the percentage of missing values, I actually got a segfault in my code. It took me a while to figure it out, but the issue seems to have to do with the score calculation of a candidate parent set, which is P(D|M) + log(|D|) / 2 * |M|, where |D| is the number of examples in your data set and |M| is the number of parameters in the model. Pomegranate handles missing data when building Bayesian networks by ignoring examples that have missing values when focusing on the subset of features corresponding to a variable and it's candidate parents. This means that |D| can change from one candidate parent set to the next. In cases where the candidate parent set is large and the data full of missing values, you may get situations where there is only one example in the data where no features are missing. In this case, P(D|M) must be 0, but log(|D|) is also 0, meaning that complexity is no longer penalized. The segfault appears to be related to this, where massive candidate sets were being selected for some variables. The fix, derived from the MDL regularization itself, appears to be to discard candidate parent sets which result in only one example, much like we discard candidate parent sets that result in zero examples.



Regardless, after incorporating this simple fix, I timed both the original code and the code incorporating this change and removing ~99% of the values. The original code:

```
Without missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.665113   8.461362
max   5.768124   8.482762
std   0.103726   0.021463

With missing values:
      cpu_time  wall_time
mean  0.171360   0.170214
max   0.172353   0.171646
std   0.000871   0.001241
```

The proposed changes:

```
Without missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.484248   8.345044
max   5.521972   8.416701
std   0.032916   0.070389

With missing values:
      cpu_time  wall_time
mean  0.160196   0.160249
max   0.161104   0.162660
std   0.001142   0.002100
```

The difference in speed between the original code and the modifications appears minor, if not just run-to-run variation. Are you sure that these changes are causing a dramatic speedup on your end? To be fair, I didn't incorporate the changes to ConditionalProbabilityTable or JointProbabilityTable because I didn't expect them to make as much of a difference as this function.

**Comment by alexhenrie** [comment]
The biggest difference actually comes from adding a shortcut to ConditionalProbabilityTable.__summarize (two underscores). Try the benchmarks again after applying the patch that touches that function.

**Comment by jmschrei** [comment]
I added in all of your changes and get the following timings when using ~99% sparsity:

```
Without missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.520043   8.370361
max   5.609389   8.449367
std   0.123448   0.125639

With missing values:
      cpu_time  wall_time
mean  0.159223   0.160261
max   0.160032   0.167227
std   0.000870   0.006134
```

When I go back to ~10% sparsity I get the following, which is comparable to the timings above:

```
Without missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.564584   8.379262
max   5.654466   8.494476
std   0.125364   0.137579

With missing values:
      cpu_time  wall_time                                                                                                                                                                
mean  5.649820   8.333919
max   5.746899   8.389962
std   0.109258   0.059623
```

**Comment by alexhenrie** [comment]
Huh. Are you sure you're running a fair comparison, making sure that Pomegranate is not installed via Pip and compiling and installing Pomegranate the same way before and after applying the changes?

If it's a fair comparison, maybe it's a difference between your Python installation, compiler, or CPU and mine. I have Python 3.7.4, GCC 9.1.0, and an Intel i5-5257U. What do you have?

**Comment by alexhenrie** [comment]
Or could your fix for data with too many missing values be affecting the performance? Could you put that fix on a branch where I can look at it?

**Comment by jmschrei** [comment]
Each time I would modify the code locally, build it, and then run the timings twice to ensure that there weren't any first-run issues. I'm running Python 3.7.3 GCC 7.3.0 in an Ubuntu vmware virtual machine on an i7-8700K 6 core (12 thread) CPU @ 3.70GHz. 

My fix involved simply adding 

```python
	if w_sum > 1:
		logp -= _log(w_sum) / 2 * m[d+1]
	else:
		logp = NEGINF
```

to the end of the `discrete_score_node` function, so I doubt that it was significantly impacting the speed of that function. It is possible that, by returning NEGINF rather than 0, that it is building a smaller network when the data has many missing values and so the time taken by `ConditionalProbabilityTable` is lessened. However, since my script segfaults with the original code, I haven't been able to time that.

alexhenrie:
I don't see any improvement when running in a virtual machine either. I think you have to be running on bare metal to see the difference.

alexhenrie:
I talked to Sergiusz and he confirmed that the bug you found could be part of our problem. However, it does not seem to be the entire problem because even with the fix, training with missing values is still significantly slower than training without missing values. But regardless of our use case, fixing the bug is the right thing to do.

alexhenrie:
Hi Jacob, it's been a week since your last comment. Do you need me to get you SSJ access to a non-VM server where you can play around and see the performance difference?

jmschrei:
The fastest way for me to provide a fix would be for me to see a data set where this is a problem. It's difficult for me to provide a fix for a bug I can't reproduce.

alexhenrie:
You should be able to reproduce the problem on the digits dataset as long as you are not running in a virtual machine. The performance characteristics of virtual machines are different from real hardware.

alexhenrie:
I guess the question is what your criteria are for accepting performance improvements. Do you only care about performance in virtual machines? If so, do you only care about VMware, or would you accept the patches if I showed that they benefit another platform such as QEMU?

jmschrei:
I ran both the current master and the proposed changes on a machine that only had Linux installed. I couldn't reproduce the gain in speed there either. The only noteworthy things about the machine is that it's part of a shared file system, and appears to be fairly slow compared to both of our machines. I don't know where this source of slowness comes from, because it allegedly has 2.4GHz processors, in comparison to my 3.7GHz ones. However, not only is the learning process slow, but it takes a few seconds to write "Without missing values" to the screen, and then another few seconds to start the progress bar...

Current master:

```
Without missing values:
       cpu_time  wall_time                                                                                           
mean  25.002199  31.330353
max   25.019196  31.929205
std    0.016997   0.518762

With missing values:
       cpu_time  wall_time                                                                                           
mean  25.142178  31.134752
max   25.156176  31.183504
std    0.013526   0.042548
```

Proposed changes:

```
Without missing values:
       cpu_time  wall_time                                                                                                                                                 
mean  25.780414  32.184617
max   25.831073  32.499963
std    0.074321   0.287220

With missing values:
       cpu_time  wall_time                                                                                                                                                 
mean  25.802744  32.714040
max   25.805077  32.910791
std    0.003214   0.170431
```

Out of curiosity, why is the progress bar out of 6 when you write `repeat=3` in the command? Also, do you know why each iteration appears to take longer than the last one? They should all be the same command on the same data, right?

My understanding of the situation is that it appears there is a bug which, on your end, is causing a major slowdown in structure learning when there are missing values. I agree with you that it is important to track down this bug and fix it. However, I haven't been able to reproduce this bug, and it doesn't look like you are either outside of this one(?) data set that you haven't shared with me. Please correct me if I'm wrong. Instead, you have proposed a few changes that, while well-intentioned, don't appear to make any difference on my end. Even when the improvements you report are taken at face value, they don't appear to solve the larger bug, and don't really appear to me to yield major speed improvements. 

I would like to see either an issue with reproducible code that demonstrates the bug so that I can begin to debug it on my end, or, even better, a PR with such code and also the fix. I would happily merge such a PR.

alexhenrie:
I posted a sample program that has similar problems to our real dataset over a month ago at https://github.com/jmschrei/pomegranate/issues/589#issuecomment-509419374. I'll respond to your other questions in the morning.

alexhenrie:
I don't know why neurtu appears to do 6 iterations with each iteration taking more time than the last. I agree that it is strange. If you prefer to use different benchmarking software, that's totally fine.

alexhenrie:
I've tried the shortcuts on several more computers now and gotten mixed results. The shortcuts seem more likely to help when the system is under load, but I haven't found a scenario where I can consistently see an improvement. I still think that merging these commits is the right thing to do primarily because they simplify the code, but I can live with it if you are still opposed.

jmschrei:
Added in a different PR.

PR #612, opened 2019-08-19
Make reduce_dataset option work with missing values

alexhenrie:
This improves running time when a lot of rows are the same due to missing values, such as in the example at https://github.com/jmschrei/pomegranate/issues/589#issuecomment-509419374.

Dataset reduction still adds overhead though, so even with these improvements it's not yet clear whether `reduce_dataset=True` or `reduce_dataset=False` will be faster for our particular data set.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -1057,7 +1057,10 @@ cdef class BayesianNetwork(GraphModel):
 			X_count = {}
 
 			for x, weight in izip(X, weights):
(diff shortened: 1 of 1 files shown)


alexhenrie:
This patch reduced the training time on our dataset to 2 hours. So now it definitely looks like the right solution :-)

jmschrei:
Awesome. Glad we could eventually sort this whole thing out. I'll get this merged soon!

jmschrei:
🎉

PR #614, opened 2019-08-27
Optimize function BayesianNetwork.discrete_exact_a_star

alexhenrie:
This patch contains four optimizations:

* Define all variable types
* Make `other_variables` a list instead of a dict
* Eliminate the variable `out_set` and instead use the iterator directly (without converting it to a tuple)
* Skip creating the `entry` variable if it's not going to be used

Individually they are not much to write home about, but together they make the function about 1% faster.

Test program:

```
import numpy as np
from random import random

```

Before optimizations:

```
      wall_time   cpu_time                                                                                                    
mean  13.072534  13.025136
max   13.097647  13.043180
std    0.024602   0.017829
```

After optimizations:

```
      wall_time   cpu_time                                                                                                    
mean  12.935084  12.884005
max   12.961816  12.912683
std    0.025821   0.029807
```

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -1599,17 +1599,18 @@ def discrete_exact_a_star(X, weights, key_count, pseudocount, max_parents, n_job
 		The parents for each variable in this SCC
 	"""
 
-	cdef int i, n = X.shape[0], d = X.shape[1]
-	cdef list parent_graphs = []
+	cdef int i, j, n = X.shape[0], d = X.shape[1]
+	cdef list parent_graphs, other_variables, prev_entry, local_structure
+	cdef tuple variables, parents, new_variables, entry
+	cdef set closed = set()
+	cdef ParentGraph pg
+	cdef double c, e, f, g, h
 
 	parent_graphs = [ParentGraph(X, weights, key_count, i, pseudocount, max_parents) for i in range(d)]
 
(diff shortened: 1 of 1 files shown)


alexhenrie:
I just added a fifth optimization:
* Make `closed` a set instead of a dict

jmschrei:
Hi Alex

As we've discussed before, I'm probably not going to be merging code modifications unless they produce a significant speed improvement. I'm not comfortable setting a specific threshold for inclusion and would rather deal with things on a case-by-case basis, but certainly a 1% speed improvement is too low without some other important improvement.

alexhenrie:
How about just making `other_variables` a list and `closed` a set, since those two changes would make the code simpler and easier to understand in addition to saving time and memory?

alexhenrie:
Jacob, do you want me to resubmit this with just the changes to `other_variables` and `closed`?

jmschrei:
I don't see what the benefit of making these changes is. In particular, a set and a dictionary both use a hash map in the Python backend, so I'm not sure there is a difference between the two for `closed`. Personally, dictionaries make more sense to me because it coincides with the idea of dictionary lookup.

alexhenrie:
Writing `closed[variables] = 1` leaves the reader wondering why the number 1 is being associated with that item. It's not immediately apparent that the number 1 is associated with _every_ item added to the set or that it could just as easily have been 0, True, False, None, or anything else.

PR #617, opened 2019-09-03
Fix copy-paste error in MultivariateDistribution.log_probability

alexhenrie:
Just a few lines above this, these variables are already defined to the same values they are being set to here.


jmschrei:
Thanks!

PR #618, opened 2019-09-03
Fix return value of Kmeans.summarize

alexhenrie:
`total_dist` was being used (via the `+=` operator) while it was still initialized to a garbage value. At the same time, `pdist` was being unnecessarily initialized at the beginning of the function only to be initialized later before it is used. So I'm guessing that you meant to put the `= 0.0` on the declaration of `total_dist` but accidentally put it on `pdist` instead.


jmschrei:
Good catch. Fortunately it seems like the only thing it could potentially effect is the number of iterations for training if not manually set.

jmschrei:
This probably also explains why k-means related unit tests would fail occasionally but be fixed by re-running them.

PR #620, opened 2019-09-04
Remove enumerate call from BayesianNetwork.generate_parent_graph

alexhenrie:
It's confusing (and time-wasting) to assign values to `k` and then not use them.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -2016,7 +2016,7 @@ def generate_parent_graph(numpy.ndarray X_ndarray,
 			else:
 				best_structure, best_score = (), NEGINF
 
-			for k, variable in enumerate(subset):
+			for variable in subset:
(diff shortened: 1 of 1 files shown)


jmschrei:
Hi Alex

I'm not going to merge this change. I don't believe it's either time wasting or confusing, even though it does end up assigning k when it doesn't need to. I thin it would be most productive if your PRs focused on substantive changes that either added in significant speed improvements, bug fixes, or new features, rather than minor tweaks to the code that you find preferable.

alexhenrie:
I apologize for repeatedly bothering you. I think you might not recognize it as a simplification because you are already intimately familiar with this code and the algorithms it implements. But now that I know that you don't consider unnecessary calls to `enumerate` to be a problem, I won't point them out to you in the future.

Look on the bright side: The more of my pull requests you reject, the better of an idea I'll have about what you do and do not consider an improvement.

PR #621, opened 2019-09-04
Fix copy-paste error in BayesModel.predict_log_proba

alexhenrie:
This is another case of repeating code that was just executed a few lines before.


jmschrei:
What lazy programmer copy/pasted a bunch of code like this?

alexhenrie:
Hahaha. Thanks Jacob!

PR #622, opened 2019-09-05
Fix calculation of h in BayesianNetwork.discrete_exact_a_star

alexhenrie:
This has got to be a typo.


jmschrei:
Good catch. Fortunately it only effects the first item pushed onto the priority queue which then gets immediately popped off. It will monotonically shift the internal log probability calculations but won't effect the resulting structure, which is why the unit tests passed.

alexhenrie:
Thank you!

PR #623, opened 2019-09-09
Fix calculation of logp in BayesianNetwork.discrete_score_node

alexhenrie:
I'm pretty sure this is another typo. Why would we clamp negative values to negative infinity?

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -2106,7 +2106,7 @@ cdef double discrete_score_node(double* X, double* weights, int* m, int* parents
 		if count > 0:
 			logp += count * _log(count / marginal_count)
 
-	if w_sum > 1:
+	if w_sum > 0:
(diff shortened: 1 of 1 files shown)


jmschrei:
The code there is what I implemented to fix the issue I mentioned where, in a data set with missing examples, you're considering a child+parent combo that only has a single example. In that situation, without that fix, the code would return a log probability of 1 for any parent combination since there's no variance among the examples. the code is a hack that says you need more than 1 example to get a real score out. `w_sum` should always be positive (and generally an integer), as it's the number of examples seen with this parent+child combo.  A more principled approach might be to calculate a maximum number of parents on the fly based on the number of examples for each child+parent combo but I haven't actually checked to see if that approach is normalized across all #s of examples or if it leads to weird structures. In contrast, this approach is both simple and solves the issue.

alexhenrie:
Okay, sure. Thanks for the explanation!

PR #624, opened 2019-09-09
Make the utils.isnan function inline

alexhenrie:
Bayes net training is about 35% faster when `isnan` is inlined, which makes sense because `npy_isnan` was designed to be an inline function.

Test program:

```
import numpy as np
from random import random

X = np.round(np.random.rand(10000, 16))
train = delayed(BayesianNetwork)().from_samples(X)

```

Non-inline `isnan`:

```
      wall_time  cpu_time                                                                                                     
mean   8.369287  8.361042
max    8.488417  8.471947
std    0.103549  0.112760
```

Inline `isnan`:

```
      wall_time  cpu_time                                                                                                     
mean   5.361331  5.360565
max    5.379823  5.376542
std    0.030915  0.019036
```

### DIFF
--- a/pomegranate/utils.pxd
+++ b/pomegranate/utils.pxd
@@ -8,7 +8,9 @@ cdef int* GPU
 cdef extern from "numpy/npy_math.h":
 	bint npy_isnan(double x) nogil
 
-cdef bint isnan(double x) nogil
+cdef inline bint isnan(double x) nogil:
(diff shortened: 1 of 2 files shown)


jmschrei:
I can reproduce this. Nice find.

alexhenrie:
Thank you!!!

Issue #548, opened 2019-02-06
pomegranate-0.10.0.tar.gz contains Cython generated .c rather than source .pyx files


mtselman:
Hi, was trying to build pomegranate from source using pomegranate-0.10.0.tar.gz from PyPi, but that tarball seems to be missing many source files while containing Cython-generated .c files.

jmschrei:
Howdy

You're getting the wheels, which are pre-built binaries that should work on your system. If you'd like the original source files, the easiest way is likely to download them directly from github.

mtselman:
Thanks for the reply!
In our environment we build everything from source [ and do not rely on pre-built binaries ]. Got the sources from github using 0.10.0 tag, but just wanted to point out that your  tar.gz files on PyPi are not usable.

jmschrei:
Okay, thanks for letting me know. I'll see if there's a quick fix for future releases.

PR #626, opened 2019-09-16
*.pxd *.pyx from pomegranate.distributions as well

mr-c:
similar to https://github.com/jmschrei/pomegranate/issues/548

This would help Debian out, thanks!


jmschrei:
Good call on both fronts. Thanks!

PR #629, opened 2019-10-02
Construct the parent subset more efficiently in ParentGraph.__getitem__

alexhenrie:
This speeds up exact Bayes net training by up to 34% when there is little to no variability in the data.

Test program:

```
import numpy as np

X = np.zeros([10000, 17])
```

Before:

```
      wall_time  cpu_time
mean   6.998546  6.838988
max    7.118689  6.925751
std    0.110920  0.098881
```

After:

```
      wall_time  cpu_time
mean   4.597856  4.502262
max    4.603789  4.547125
std    0.005501  0.048928
```

The speed improvement diminishes as variability in the training dataset increases. If 5% of the zeroes are replaced with ones at random, the speed improvement drops to about 17%, and if 10% of the zeroes are replaced with ones at random, the speed improvement drops to about 11%.

I checked to make sure that the resulting Bayes net is the same with or without this change.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -1218,8 +1218,8 @@ cdef class ParentGraph(object):
 		else:
 			best_parents, best_score = value, self.calculate_value(value)
 
-		for variable in value:
-			parent_subset = tuple(parent for parent in value if parent != variable)
+		for i in range(len(value)):
+			parent_subset = value[:i] + value[i+1:]
 			parents, score = self[parent_subset]
 
 			if score > best_score:


jmschrei:
I don't understand how this could lead to that much of an improvement. Going through and calculating the log likelihoods for each parent set should take much more time than simply constructing a list minus one item. Can you provide some insight into why it helps?

jmschrei:
Thinking about this more, this is likely also related to the data set reduction step. When the entries are entirely 0, the data set gets reduced to 1 example. In this case, calculating the log probability is fairly fast. When a few of the values are set to 1, the data set gets reduced to a small number of examples. The log probability calculation takes longer, but not that much longer. 

I will try reproducing this on my side soon. I have a paper deadline on Friday so I can't promise I'll get to it before then, but I'll try to get to it soon. I'll probably also release a new version after incorporating this, unless you find something new before then.

PR #635, opened 2019-10-15
Suggest replacing MDL with BIC for Structure Learning documentation

BKrajancic:
The original definition of the MDL score is to use a log with base 2, however pomegranate uses a base of 'e' in all of its calculations.  This may present an issue when comparing to software which uses log_2.

It can be argued that as calculation of scores is positive rather than negative it is the MDL. However I believe that the consideration of floating point error when calculating with log_2 rather than log_e is more important to detail, as structure learning which uses log_2 may produce different results.

An alternative solution is to document the base of the log used.

Original Definition is from:
http://www.cs.toronto.edu/~fbacchus/Papers/pricai.pdf


jmschrei:
It is a reasonable point that we use base e rather than base 2 and that is worth noting in the documentation. However, I don't think that using base e makes it the Bayesian information criterion. Can you explain why you think that it is?

BKrajancic:
I'm on mobile and I'll give a proper response when I can. Other software such as BNLearn use Base E and call it the BIC. There's also the wikipedia page which in the properties section references similarity to MDL. 
https://en.m.wikipedia.org/wiki/Bayesian_information_criterion

jmschrei:
Okay. Well, I'm certainly open to clarifying that we use base e in the documentation and suggesting that this is similar (or equivalent if we can find some references) to the BIC. Thanks for catching this!

BKrajancic:
I'll post more tonight, but a good start is page 2156 http://www.jmlr.org/papers/volume7/decampos06a/decampos06a.pdf&ved=2ahUKEwj4loOAxPDlAhX97XMBHYpOAPQQFjABegQIBRAC&usg=AOvVaw31Q9BkrqL0nIO3S7p8r_jN

"scoring function for learning bayesian networks
based on mutual information and conditional independence tests"

BKrajancic:
In the article "Learning Bayesian Belief Networks
An approach based on the MDL Principle" which is the first paper to use MDL for learning Bayesian Networks, every log base is 2. 

A fantastic source which describes both scores as being equivalent is:
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3439716/#!po=30.2885 

I will endeavour to find sources which define BIC as being ln.

Hopefully these make my argument clearer for changing the definition.

BKrajancic:
Hi, i'm finally sitting down at a computer to give this topic a proper conclusion from my perspective.

The original paper which presents the BIC does not state which logarithm base is used:
https://projecteuclid.org/download/pdf_1/euclid.aos/1176344136

However Scharwz's paper presents a scoring criteria which applies a penalty to Akaike's score which was published in the paper:
http://bayes.acs.unt.edu:8083/BayesContent/class/Jon/MiscDocs/Akaike_1974.pdf and within this paper "where, as in the sequel of the present paper, log denotes the natural logarithms."

I realise now I didn't give a link to the paper "A Scoring Function for Learning Bayesian Networks based on Mutual Information and Conditional Independence Tests" 
http://www.jmlr.org/papers/volume7/decampos06a/decampos06a.pdf

The paper generalises AIC, BIC and MDL in equation 5, and the equation shows how the function performs in logarithmic space (making log_e and log_2 equivalent). Although it is incorrect to argue that BIC must use the natural logarithm, it is clear that there is no strict definition and any base can be used in order for a score to be BIC. Furthermore it is clear that MDL must use log_2, based on its original publication and the principle of MDL (a number's minimum description length being based on the number of bits to represent each number).

Please let me know if there is any more information I need to provide on this topic.

jmschrei:
Thanks for taking the time to look into this. It is on my to-do list, but the change probably won't make its way into the next version I'm releasing this week because I'm fairly strapped for time. I haven't forgotten about this though.

jmschrei:
I got around to looking into it more. I've changed the internals to explicitly use base2 instead of base e but there won't be a difference in result because it's just a rescaling issue.

PR #643, opened 2019-10-22
Filter out NaNs in the ParentGraph constructor

alexhenrie:
This makes the exact training algorithm about 10% faster when there are no missing values and about 75% faster when 10% of the values are missing.

Test program:

```
import numpy as np
from random import random

print()

for i in range(X.shape[0]):
    for j in range(X.shape[1]):
        if random() <= 0.1:
            X[i][j] = np.nan

print('With missing values:')
print()
```

Before:

```
Without missing values:
      wall_time  cpu_time                                                                                                     
mean   7.898006  7.920488
max    7.947955  8.061725
std    0.043266  0.122328

With missing values:
      wall_time   cpu_time                                                                                                    
mean  13.262545  13.249452
max   13.285537  13.286770
std    0.031434   0.032632
```

After:

```
Without missing values:
      wall_time  cpu_time                                                                                                     
mean   7.136354  7.082058
max    7.185820  7.094680
std    0.042866  0.010932

With missing values:
      wall_time  cpu_time                                                                                                     
mean   3.139106  3.120517
max    3.148663  3.145033
std    0.008417  0.021324
```

I checked and the resulting Bayes net is exactly the same before and after the optimization.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -154,6 +154,9 @@ def _check_input(X, model):
 
 	return True
 
+cdef numpy.ndarray exclude_missing(numpy.ndarray X):
+	return X[numpy.logical_not(numpy.isnan(X).any(axis=1))]
+
 
 cdef class BayesianNetwork(GraphModel):
 	"""A Bayesian Network Model.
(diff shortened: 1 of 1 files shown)


jmschrei:
Currently, pomegranate will derive model parameters using the present
aspects of any examples, e.g. the probabilities in conditional probability
tables when the respective variables are present but others are missing.
This takes the most advantage of the given data. It looks like this
proposal is to eliminate any example with one or more missing values. While
it will naturally speed things up, it won’t actually result in the same
networks because a bunch of the information is discarded.

alexhenrie:
Ohh I see. `discrete_score_node` only ignores examples where the _parents_ are not missing (it's fine if other values are missing). Thanks for the clarification.

PR #644, opened 2019-10-22
Keep a pointer to the current row instead of recalculating its location

alexhenrie:
This makes the exact Bayes net training algorithm about 5% faster. Plus, the new variable makes the code easier to follow.

Test program:

```
import numpy as np

```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   7.731686  7.716043
max    7.751758  7.718887
std    0.019498  0.004410
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   7.444296  7.378512
max    7.472569  7.391443
std    0.024586  0.011575
```


jmschrei:
I can reproduce this. It looks closer to 10% on my machine. Good catch. However, "sample" is a very overloaded term and so I generally prefer either "row" or "example," which I think are clearer. I'm going to change it to be "row."

alexhenrie:
If you prefer, I can rebase this patch and change "sample" to "row" myself. It would be a lot cleaner than having a merge commit plus two commits that change things.

jmschrei:
Squashing the commits should do the trick. Thanks again for the suggestion.

PR #648, opened 2019-11-05
Store Chow-Liu structure in lists while it is being built

alexhenrie:
This decreases the time it takes to train a Chow-Liu net by about 5%. (Tuples, being immutable, are expensive to manipulate.)

Test program:

```
import numpy as np
from random import random

X = np.round(np.random.rand(10000, 100))
train = delayed(BayesianNetwork)().from_samples(X, algorithm='chow-liu')

```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   9.900260  9.883265
max    9.965771  9.914445
std    0.073636  0.037711
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   9.445535  9.361896
max    9.481115  9.454669
std    0.052243  0.080365
```


jmschrei:
Good catch, thanks!

PR #652, opened 2019-11-13
Refactored and fixed checking NaNs

mareksmid-lucid:
Sometimes BayesianNetwork ignored a NaN value in the input when trained from samples, because `DiscreteDistribution._log_probability` checked only for Python floats, not also NumPy floats.
Also did some refactoring.


jmschrei:
This is great, thanks!

PR #655, opened 2019-11-14
Remove unnecessary calls to memset after calloc

alexhenrie:
The calloc function is guaranteed to zero out the memory before returning a pointer to it.


jmschrei:
Great, thanks!

PR #657, opened 2019-11-20
Deprecation warning YAMLLoadWarning: Add Loader argument for yaml loading

m-martin-j:
Small change made, in order to deal with deprecation warning: YAMLLoadWarning, when calling from_yaml e.g. for an HMM instantiation from yaml string
"default loader is unsafe" so I offer to replace it by yaml.Loader,
(see also https://msg.pyyaml.org/load)
(one of my first pull requests, plz be gentle =))

### DIFF
--- a/pomegranate/base.pyx
+++ b/pomegranate/base.pyx
@@ -70,7 +70,7 @@ cdef class Model(object):
 	@classmethod
 	def from_yaml( cls, yml ):
 		"""Deserialize this object from its YAML representation."""
-		return cls.from_json(json.dumps(yaml.load(yml, Loader=yaml.Loader)))
+		return cls.from_json(json.dumps(yaml.load(yml, Loader=yaml.SafeLoader)))
 
 	def copy(self):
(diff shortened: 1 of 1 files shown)


m-martin-j:
Haven't executed tests but it checks out for instantiating an hmm from yaml

jmschrei:
Thanks for catching this. Can you test whether the SafeLoader would work as well? Might as well improve the safety if it doesn't effect performance.

m-martin-j:
Reasonable though; done. Using yaml.SafeLoader took only 0.00021 longer compared to using yaml.Loader on my i7-6700 (for an hmm from the first 6 cells of [this example](https://github.com/jmschrei/pomegranate/blob/master/tutorials/B_Model_Tutorial_3_Hidden_Markov_Models.ipynb)).

Tests ran through, too.

```
----------------------------------------------------------------------
Ran 457 tests in 16.364s

OK
```

jmschrei:
Thanks again!

m-martin-j:
Pleasure

PR #658, opened 2019-11-22
Include LICENSE file

toddrme2178:
Include the LICENSE file, as the MIT license requires it be included in all copies of the software.  Also simplify the `MANIFEST.in` file a little.


jmschrei:
Good addition, thanks!

PR #662, opened 2019-11-25
Fix CI badges

XVilka:
Noticed at readthedocs: https://pomegranate.readthedocs.io/en/latest/


jmschrei:
Good call. Thanks.

XVilka:
@jmschrei interestingly, it wasn't update at ReadTheDocs: https://pomegranate.readthedocs.io/en/latest/

Do you need to refresh it somehow?

PR #665, opened 2019-11-29
Update requirements.txt

adrien-mogenet:
Added matplotlib, pillow and pygraphviz as required so that one can use the plot functions from Pomegranate.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,6 @@ joblib >= 0.9.0b4
 networkx >= 2.0
 scipy >= 0.17.0
 pyyaml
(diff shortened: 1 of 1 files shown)


adrien-mogenet:
Alternative: update the tutorials or the README file.

jmschrei:
Thanks for the suggestion. I'd rather not make matplotlib or pygraphviz a dependency for pomegranate because that can make it more difficult to install. Because plotting is not an essential part of pomegranate, I'd rather keep it as-is, where you can download these libraries separately if you'd like.

PR #667, opened 2019-12-05
Get rid of mutual_info_ndarray

alexhenrie:
This change just makes the code a little simpler (it doesn't noticeably improve performance).


jmschrei:
Great. Now that I got rid of using it for the second part it makes sense to get rid of it entirely. Can you add freeing the array at the end of the functions so we don't have memory leaks?

alexhenrie:
Oh duh. I just added the calls to free.

PR #670, opened 2020-01-02
Relax a few tests due to a very small rounding difference

mr-c:
This fixes the i386 build on Debian.


jmschrei:
Thanks for the contribution!

PR #674, opened 2020-01-10
Fix notebooks to work with python3 

ksachdeva:


jmschrei:
Howdy. Thanks for the PR. Can you describe what it is you did to the notebooks?

ksachdeva:
Hi @jmschrei 

The modifications are -

a) Convert python 2 'print' statements to python 3 i.e. but brackets. 

b) Few notebooks did not have proper float to int conversions leading to indexing problems. So there I used `//` instead of '/'

c) matplotlib does not take 'map' as the input so I had to pass the result of map to list before supplying it to a matplotlib function

I have listed these in the revision comment as well (see individual revision commit message). In the examples notebook only change to print was required

Regards
Kapil

jmschrei:
I looked over a few of them and they look good so I'm going to merge. Thanks for the contribution!

PR #703, opened 2020-03-10
Define GPU and has_cupy as bints

alexhenrie:
This both makes the code easier to read and avoids using calloc without a matching call to free.

### DIFF
--- a/pomegranate/utils.pxd
+++ b/pomegranate/utils.pxd
@@ -3,15 +3,13 @@
 
 cimport numpy
 
-cdef int* GPU
-
 cdef extern from "numpy/npy_math.h":
 	bint npy_isnan(double x) nogil
 
 cdef inline bint isnan(double x) nogil:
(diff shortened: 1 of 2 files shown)


jmschrei:
Are you sure this works? The reason I coded it like this is because I had trouble accessing global variables without making them arrays like this.

alexhenrie:
I'm not having any trouble at all with the non-pointer global variables:

```
$ python
Python 3.8.2 (default, Feb 26 2020, 22:21:03) 
[GCC 9.2.1 20200130] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pomegranate import utils
>>> utils.is_gpu_enabled()
True
>>> utils.disable_gpu()
>>> utils.is_gpu_enabled()
False
>>> utils.enable_gpu()
>>> utils.is_gpu_enabled()
True
```

PR #704, opened 2020-03-11
Drop support for Python 2

alexhenrie:
No one should be using Python 2 anymore, and creating aliases for the range and zip functions prevents the Cython compiler from replacing those functions with optimized C code.

When putting parentheses around the arguments to the print function, I have preserved the extra spaces in files that use that convention when calling functions, and left them out in files that generally do not use that convention.

I have not attempted to update any of the Jupyter notebooks.


jmschrei:
Thanks. I had actually already removed Python 2 support from most of the package but missed this.

PR #705, opened 2020-03-12
Fix math and json imports in hmm.pyx

alexhenrie:
Currently HiddenMarkovModel.to_json is totally broken because the modules it depends on are not imported correctly.


jmschrei:
Wow. How did that happen???

jmschrei:
I'm not sure that it actually broke HMM serialization because I have unit tests to make sure that works, and have personally used it in the past. I'm guessing that those names got overriden somewhere later on. It's def incorrect though.

PR #709, opened 2020-03-15
Remove unused imports

alexhenrie:
This primarily makes the code easier to understand, but it can also reduce the time it takes for scripts to start up as well as reducing the amount of memory they consume.


jmschrei:
Would you mind resolving the conflicts here and resubmitting? I'd be happy to merge it then.

alexhenrie:
Done.

PR #723, opened 2020-04-09
More calloc and memset cleanup

alexhenrie:
Somehow I missed these in https://github.com/jmschrei/pomegranate/pull/655


alexhenrie:
Thank you!

PR #727, opened 2020-04-22
Reimplement DiscreteDistribution.__imul__

alexhenrie:
Avoiding unnecessary reinitialization decreases the time needed to make predictions by about 50%.

Test program:

```
import numpy as np
from random import random

```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   0.002891  0.002878
max    0.002938  0.002927
std    0.000013  0.000011
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   0.001496  0.001493
max    0.001551  0.001557
std    0.000011  0.000014
```

I took care to keep the behavior identical to the original, but even more time could be saved if the summaries etc. do not need to be reset.

I also threw in a small cleanup that simplifies getting the data type name in the DiscreteDistribution constructor and a couple of other places. Let me know if you'd rather review that in a separate pull request though.

### DIFF
--- a/pomegranate/distributions/ConditionalProbabilityTable.pyx
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pyx
@@ -55,7 +55,7 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 
 		self.dtypes = []
 		for column in table[0]:
-			dtype = str(type(column)).split()[-1].strip('>').strip("'")
+			dtype = type(column).__name__
 			self.dtypes.append(dtype)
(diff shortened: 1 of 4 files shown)


alexhenrie:
Sorry, this needs a bit more work. I'll ping you when I've fixed the test failures.

alexhenrie:
Okay, after fixing the bugs there doesn't appear to actually be any noticeable speed improvement to reimplementing `__imul__`, so I am closing this pull request and will instead open a new one with just the type name cleanup.

PR #728, opened 2020-04-22
Use type.__name__ instead of str(type) and string manipulation

alexhenrie:
The code is a lot more understandable this way.

### DIFF
--- a/pomegranate/distributions/ConditionalProbabilityTable.pyx
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pyx
@@ -56,7 +56,7 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 
 		self.dtypes = []
 		for column in table[0]:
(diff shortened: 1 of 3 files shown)


alexhenrie:
Thanks! By the way, just after you merged this pull request, I realized that this commit changes `"dtype" : "numpy.float64"` to simply `"dtype" : "float64"` in JSON files. It looks okay to me to omit the numpy prefix, but if you can think of a reason to keep including it then maybe the commit should be reverted.

jmschrei:
Where are you seeing a situation where it would write out `"dtype" : "numpy.float64"`?

alexhenrie:
For example:

```

X = load_digits(return_X_y=True)[0][:,:5]
print(BayesianNetwork().from_samples(X, algorithm='exact').to_json())
```

It's really embarrassing that I didn't notice this earlier. The commit should definitely be reverted.

PR #729, opened 2020-04-23
Don't use numpy.zeros for memory that is about to be reinitialized

alexhenrie:
In the same vein as #655 and #723, this commit speeds up various functions by eliminating unnecessary memory operations.


alexhenrie:
Thank you!

PR #730, opened 2020-04-23
Revert "Use type.__name__ instead of str(type) and string manipulation (#728)"

alexhenrie:
This reverts commit f5f387c8f2ecf6ef5c2be74764b479194ebae4a6.

That was a mistake, sorry. `Distribution.from_json` depends on the dtype starting with `numpy.` if it is a numpy type.


jmschrei:
Oops. No problem. I agree that your way reads nicer and is more intuitive. Happy to look at another PR if you can find a way around this problem. If you do submit another PR can it include a unit test that checks for this issue?

alexhenrie:
I just made some tests to catch this kind of error and I will send them in a new pull request.

PR #733, opened 2020-04-23
Skip convergence checks after the first failed check

alexhenrie:
This is a simple change that decreases the time required to make Bayes net predictions by about 7%.

Test program:

```

```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   0.002968  0.002968
max    0.003024  0.003377
std    0.000013  0.000060
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   0.002737  0.002729
max    0.002787  0.002764
std    0.000018  0.000012
```

### DIFF
--- a/pomegranate/FactorGraph.pyx
+++ b/pomegranate/FactorGraph.pyx
@@ -275,6 +275,7 @@ cdef class FactorGraph(GraphModel):
 		#   (2) send messages from the factors to the variables, containing
 		#   the factors belief about each marginal.
 		# This is the flooding message schedule for loopy belief propagation.
+		cdef bint done
 		iteration = 0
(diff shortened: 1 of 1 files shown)


jmschrei:
Thanks for the contribution. Wouldn't it make more sense to change the loop to terminate once it found a distribution that hadn't converged?

alexhenrie:
It looks like this code depends on all of the distributions being updated before the next pass in order to determine whether or not the next pass converges. That would mean that the output could change if the loop that updates the distributions is exited early. But you're the expert on these algorithms, so if there's some reason why exiting early can't change the output or delay convergence, I would be all for it.

PR #736, opened 2020-04-24
Remove unused variables from JointProbabilityTable class

alexhenrie:
This commit removes code that was unnecessarily copied and pasted from ConditionalProbabilityTable to JointProbabilityTable, which results in cleaner code, less memory usage, and a slight speed improvement to Bayes net predictions.


alexhenrie:
Thank you!

Issue #400, opened 2018-03-19
Hidden Markov model with Gaussian emissions not training properly


joshdorrington:
I am trying to fit a time series of a chaotic system with a 3 state hidden markov model with gaussian emissions. I format the data as a numpy array called input_data, and run the code below:

`
```
In[1]: input_data.shape
Out[1]: (1, 200001, 6)
In[2]: model=HiddenMarkovModel()
In[3]: gaussmodel=model.from_samples(MultivariateGaussianDistribution, n_components=3, X=input_data)
```

`

Unfortunately the returned states of the model are not accurate, all components of the gaussian means are nearly identical and strictly descending absolute value:

`
```
In[4]: gaussmodel.states[1].distribution.parameters[0]
Out[4]: 
[-0.09880035589313062,
 -0.0979069486147086,
 -0.09692864738743179,
 -0.09590506750933378,
 -0.09484837637195158,
 -0.09376959974296883
```

```
In[5]: gaussmodel.states[0].distribution.parameters[0]
Out[5]: 
[0.25390719830239683,
 0.2529250438880774,
 0.25183304658880706,
 0.25062083419641734,
 0.24927706774681227,
 0.24778986004254214]`
```

The data set has no such symmetry to it, and in fact these mean values are in some cases totally outside the range of values present in the data, so I do not think it is a problem with the kmeans initialisation of the states.

jmschrei:
Howdy

It would be easiest for me to debug if you could provide the data as well. Can you also check the version of pomegranate that you have?  Taking a look at the results when you toggle `verbose=True` might help as well.

joshdorrington:
I'm on 0.7.3.
Will try and reduce the data a bit for upload, it is quite a large file at present.

jmschrei:
Can you try also upgrading to 0.9.0 as well?

joshdorrington:
The data file is attached:
[np_input_file.txt](https://github.com/jmschrei/pomegranate/files/1829676/np_input_file.txt)

The error persists unchanged with 0.9.0


```
data=np.fromfile("np_input_file.txt").reshape([1,200001,6])
model=HiddenMarkovModel()
Gaussmodel=model.from_samples(MultivariateGaussianDistribution, n_components=3, X=data)

Gaussmodel.states[1].distribution.parameters[0]
```
We see that when the data and regime centroids are projected onto the 0th and 3rd axes, there is no relation to the region the points lie in:

[...]

PR #751, opened 2020-05-11
Check that the JSON dtype is a numpy type before treating it as such

alexhenrie:
It's best to be explicit about this so that other namespaced types can be supported in the future.

### DIFF
--- a/pomegranate/distributions/distributions.pyx
+++ b/pomegranate/distributions/distributions.pyx
@@ -299,8 +299,8 @@ cdef class Distribution(Model):
 				dist = {int(key) : value for key, value in dp.items()}
 			elif d['dtype'] == 'float':
 				dist = {float(key) : value for key, value in dp.items()}
-			elif '.' in d['dtype']:
-				dtype = d['dtype'].split('.')[-1]
+			elif d['dtype'].startswith('numpy.'):
+				dtype = d['dtype'][6:]
 				dist = {numpy.array([key], dtype=dtype)[0]: value for key, value in dp.items()}
 			else:
 				dist = dp
@@ -322,8 +322,8 @@ cdef class Distribution(Model):
 						table[-1].append(int(item))
 					elif dtype == 'float':
 						table[-1].append(float(item))
-					elif '.' in dtype:
-						dtype = dtype.split('.')[-1]
+					elif dtype.startswith('numpy.'):
+						dtype = dtype[6:]
 						table[-1].append(numpy.array([item], dtype=dtype)[0])
 					else:
 						table[-1].append(item)


pomegranate/distributions/distributions.pyx
```diff
@@ -299,8 +299,8 @@ cdef class Distribution(Model):
 				dist = {int(key) : value for key, value in dp.items()}
 			elif d['dtype'] == 'float':
 				dist = {float(key) : value for key, value in dp.items()}
-			elif '.' in d['dtype']:
-				dtype = d['dtype'].split('.')[-1]
+			elif d['dtype'].startswith('numpy.'):
```
jkleckner, line 302: @alexhenrie You have done a lot of nice cleanups and checks.  Where might one put a check on C ordering for numpy objects?  That as been a recurring issue for repeatability.

alexhenrie: I'm not sure what you mean. This function only uses 1-dimensional numpy arrays, so there is no difference between C and Fortran array orderings. In what situation are you seeing an ordering mixup?

jkleckner: @alexhenrie I was prompted by this PR from @MichaelMarien [1] to think about the ordering issue.  I've never dug into it deeply.  I wonder what happens if someone passes in a 1D slice from a 3D array?  I found myself in the past having to insert a `copy()` after doing a transpose such as `d.fit(s.values.T.copy())` otherwise I saw strange results as in this conversation [2] and this conversation [3].

[1] https://github.com/jmschrei/pomegranate/pull/740
[2] https://github.com/jmschrei/pomegranate/issues/414#issuecomment-390734369
[3] https://github.com/jmschrei/pomegranate/issues/400#issuecomment-375742716

jkleckner: I a rewrite of my question would be "Is there a check for row-major ordering that might make sense in some place like this?"

alexhenrie: Array ordering can only be a problem when a multidimensional array is passed to Pomegranate by an external program and Pomegranate takes a pointer to the array data without checking the array order. I looked through the code and the only places where I believe that that can still happen are all in the k-means code. I have sent pull request https://github.com/jmschrei/pomegranate/pull/753 to address the k-means problems.

alexhenrie: To answer your question about what happens if a noncontiguous array slice is passed in:

```
>>> numpy.array([[1,2],[3,4]])[:,1].flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

>>> numpy.array(numpy.array([[1,2],[3,4]])[:,1]).flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

As you can see, when the noncontiguous array is passed to the numpy.array function, it is always converted to a contiguous array.

jkleckner: Awesome, thanks!

PR #753, opened 2020-05-12
Ensure that multidimensional k-means input arrays are in C order

alexhenrie:
This is a follow-up to commit cafeee45bec517d479e2579e6ee5169eb94b8103.


jkleckner:
@alexhenrie You are so prolific!  Very cool, thanks.

PR #759, opened 2020-05-19
Use regular dict instead of OrderedDict

alexhenrie:
This is really just a code cleanup: OrderedDict is deprecated because Python dicts are guaranteed to be ordered from Python 3.5 on. For what it's worth though, the switch does make Bayes net predictions 1-3% faster.

### DIFF
--- a/pomegranate/distributions/ConditionalProbabilityTable.pxd
+++ b/pomegranate/distributions/ConditionalProbabilityTable.pxd
@@ -17,8 +17,8 @@ cdef class ConditionalProbabilityTable(MultivariateDistribution):
 	cdef public numpy.ndarray column_idxs
 	cdef int* column_idxs_ptr
 	cdef public list parents, parameters, dtypes
-	cdef public object keymap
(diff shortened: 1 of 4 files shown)


jmschrei:
Can you point me to where it says that Python dicts are ordered in 3.5? What I found says 3.6 on. I'm somewhat reluctant to make this change because, apparently, there might be inconsistencies between IPython and Python about how to handle OrderedDicts. http://gandenberger.org/2018/03/10/ordered-dicts-vs-ordereddict/ Do you have thoughts on that?

alexhenrie:
You're right, the behavior changed in 3.6 and wasn't guaranteed until 3.7. However, Python 3.5 is no longer supported and even extremely stable Linux distributions like CentOS have moved to 3.6.

The IPython bugs related to dict ordering were fixed in IPython 5.6 and 6.3 (see https://github.com/ipython/ipython/pull/10959) so we don't have to worry about IPython either.

PR #760, opened 2020-05-20
Cast everything to str instead of doing type checking in _check_nan

alexhenrie:
It's a bit counterintuitive, but this change makes Bayes net predictions about 30% faster.

Test program:

```

```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   0.002025  0.002019
max    0.002062  0.002119
std    0.000007  0.000016
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   0.001402  0.001397
max    0.001441  0.001410
std    0.000008  0.000005
```

### DIFF
--- a/pomegranate/utils.pyx
+++ b/pomegranate/utils.pyx
@@ -450,11 +450,7 @@ def weight_set(items, weights):
 
 def _check_nan(X):
 	"""Checks to see if a value is nan, either as a float or a string."""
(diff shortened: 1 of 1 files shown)


alexhenrie:
Actually, it's even faster if you keep the type check but use the C isnan function instead of the numpy one. I'll make a new pull request.

PR #763, opened 2020-05-22
Move assertion after shortcut in DiscreteDistribution.__mul__

alexhenrie:
This patch makes `DiscreteDistribution.__mul__` look more like `DiscreteDistribution.equals`, which helps make clear that the same shortcut logic is being used in both functions. Plus, when combined with the [isnan improvement I sent earlier](https://github.com/jmschrei/pomegranate/pull/761), moving the assertion improves performance by 1-2% when making predictions from a Bayes net.

Test program:

```

```

With isnan improvement only:

```
      wall_time  cpu_time                                                                                                     
mean   0.001354  0.001353
max    0.001400  0.001372
std    0.000008  0.000006
```

With isnan improvement and the assertion moved after the shortcut:

```
      wall_time  cpu_time                                                                                                     
mean   0.001332  0.001330
max    0.001348  0.001344
std    0.000006  0.000005
```

### DIFF
--- a/pomegranate/distributions/DiscreteDistribution.pyx
+++ b/pomegranate/distributions/DiscreteDistribution.pyx
@@ -80,7 +80,6 @@ cdef class DiscreteDistribution(Distribution):
 
 		self_keys = self.keys()
 		other_keys = other.keys()
-		assert set(self_keys) == set(other_keys)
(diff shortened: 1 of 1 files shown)


alexhenrie:
Thank you!

PR #764, opened 2020-05-26
Move assignment of in messages for marginal nodes outside of loop

alexhenrie:
Moving this assignment out of the loop makes it clear that `in_messages[i]` is being set and not `in_messages[k]`.


jmschrei:
This is actually a bug. That line should be `in_messages[k]` but because those indexes never get read it doesn't matter. I'm going to just remove the line.

alexhenrie:
Great, thanks!

PR #765, opened 2020-05-26
Avoid using numpy slices in FactorGraph.predict_proba

alexhenrie:
This reduces the time required to make Bayes net predictions by about 2%.

Test program:

```

```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   0.001286  0.001286
max    0.001304  0.001318
std    0.000005  0.000006
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   0.001258  0.001255
max    0.001276  0.001268
std    0.000006  0.000004
```

### DIFF
--- a/pomegranate/FactorGraph.pyx
+++ b/pomegranate/FactorGraph.pyx
@@ -371,16 +371,20 @@ cdef class FactorGraph(GraphModel):
 			iteration += 1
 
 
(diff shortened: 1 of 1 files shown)


jmschrei:
In general I'm still reluctant to add very small performance improvements but I do agree that this makes the code more readable.

PR #778, opened 2020-06-17
Bayesian network sampling 

pascal-schetelat:
Add 2 sampling methods for Bayesian networks : 

- Rejection sampling successively sample each node given its parents evidence. When evidences are given on	non-root nodes, only draws that are compatible with evidence nodes are not rejected. Rejection sampling is a good option when evidences nodes are not far from the root nodes or when given evidence is likely. Rare evidences 	lead to a high rate of rejected samples, thus to significant slow down of the sampling.

- Gibbs sampling scheme is a Markov Chain Monte Carlo (MCMC) technique designed to speed up the sampling. It builds conditional probability of state transition of each nodes given its neighbours in its markov blanket.  Works well with a lot of evidences in the network, even when they are far from the root nodes. Drawback : 	convergence is only guaranteed when there is a non null probability path between states. If the posterior consists of isolated islands of high probability, Gibbs sampling will stay stuck in one the island and will never transition to the others. Successive samples will have high correlation.


pascal-schetelat:
I tried to rebase my commits but being not very familiar with the procedure, I hope I did not caused havoc.

I think there is some discussion to be had with the API and how we pass evidences and get outputs.

I also introduiced 2 very simplistic tests. rejection sampling relies on discreteDistribution.sample method for its implementation, and as such play well with random_state seed.

I made something more involved for the gibbs sampler. I implemented a random_choice (choose_one) in utils.pyx in cython in order to gain a massive boost in performance.

But I did not implemented seeding in this one yet. Hence, Gibbs samling test might randomly fails sometimes when the predicted proportion is too far from the reference.

pascal-schetelat:
Hi there, 

Just to let you know we added a seed in the sampling method (some part were still not seeded).
We also fixed a broken test (test_distribution) due to a missing seed that is not related to this pull request (but broke the test suite).

Do you need any help / information in order to merge this pull request ?

jmschrei:
Sorry, why are the tests failing? It also looks like Travis isn't running on this PR?

nicods-cstb:
Hello,
I had to fix the dev tree in our git fork, but now every check passes.
Do you have any more requirements for accepting this merge request ?
Greetings

jmschrei:
Hi. As I asked before, why aren't the travis checks running?

jmschrei:
Thanks for the contribution!

PR #792, opened 2020-07-13
Eliminate unnecessary variables X_int and X_int_ptr

alexhenrie:
In MarkovNetwork, `X_int` is not used at all and does not appear to be in any way necessary to calculate.

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -727,15 +727,14 @@ cdef class BayesianNetwork(GraphModel):
 		indices = {state.distribution: i for i, state in enumerate(self.states)}
 
 		n, d = len(X), len(X[0])
-		cdef numpy.ndarray X_int = numpy.empty((n, d), dtype='float64')
-		cdef double* X_int_ptr = <double*> X_int.data
+		cdef double* X_int = <double*> malloc(n * d * sizeof(double))
 
 		for i in range(n):
 			for j in range(d):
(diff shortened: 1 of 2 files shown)


jmschrei:
When using `nogil` you need to pass a pointer to the data rather than the numpy array itself, which can cause problems. I agree that the code block isn't used for MarkovNetwork, so I've removed it on my end. Thanks!

alexhenrie:
> When using `nogil` you need to pass a pointer to the data rather than the numpy array itself, which can cause problems.

I don't know what you mean by this comment. The point of the change was to not use numpy at all for this variable. `X_int` will remain allocated for the entire duration of the `with nogil` section; it cannot be deleted until the function completes. What am I missing?

jmschrei:
You changed `(<Model> self.distributions_ptr[i])._summarize(X_int_ptr, weights_ptr, n, 0, 1)` to `(<Model> self.distributions_ptr[i])._summarize(X_int, weights_ptr, n, 0, 1)`. I'm saying that the nogil function expects a C pointer, not a numpy array.

alexhenrie:
Right. I also changed `X_int` from a numpy array to a C pointer.

jmschrei:
Oh, I missed that. Well-- I already made the MarkovNetwork change on my end, which seems like the bigger contribution. Sorry about the confusion.

alexhenrie:
I'll rebase this and open a new pull request.

PR #793, opened 2020-07-13
Add (from|to)_dict methods for fast, reusable (de)serialization

alexhenrie:
This decreases the time required to load a 5 MB Bayes net from JSON by about 33%, while at the same time it cleans up the code by factoring out JSON serialization and deserialization logic that is common to all models.

Test program:

```
from neurtu import Benchmark, delayed

json = open('test.json').read()
load_net = delayed(BayesianNetwork.from_json)(json)
print(Benchmark(wall_time=True, cpu_time=True, repeat=100)(load_net))
```

Before:

```
      wall_time  cpu_time                                                                                                     
mean   0.296894  0.294889
max    0.303662  0.299657
std    0.002196  0.002272
```

After:

```
      wall_time  cpu_time                                                                                                     
mean   0.198973  0.196935
max    0.220578  0.203941
std    0.002785  0.002416
```

### DIFF
--- a/pomegranate/BayesClassifier.pyx
+++ b/pomegranate/BayesClassifier.pyx
@@ -3,7 +3,6 @@
 # BayesClassifier.pyx
 # Contact: Jacob Schreiber ( jmschreiber91@gmail.com )
 
-import json
 import numpy
 cimport numpy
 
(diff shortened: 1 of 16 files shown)


jmschrei:
Great, thank you!

alexhenrie:
Thank you!

Issue #786, opened 2020-07-05
[BUG] Incorrect docstring examples within BayesianNetwork module


minelminel:
- Package: `pomegranate==0.13.3`
- Source: PyPI
- OS: `Ubuntu 18.04.4 LTS`

As discussed in [#137](https://github.com/jmschrei/pomegranate/issues/136#issuecomment-651255799) there are some inconsistencies between updated package implementations and the corresponding docstring examples. To the best of my judgement, all the included Jupyter Notebook tutorials & examples are demonstrated correctly; however, to those of us using IDE docstring previews this can cause some real head scratchers.

### 1. `BayesianNetwork.add_nodes` no longer accepts a list, rather the elements should be unpacked

`BayesianNetwork.pyx` _L168_
```python
# current
>>> model.add_nodes([s1, s2])
# fixed
>>> model.add_nodes(s1, s2)
```

### 2. `BayesianNetwork.log_probability` & `BayesianNetwork.probability` should be given a *list of lists* or `numpy.ndarray`

`BayesianNetwork.pyx` _L171_
```python
# current
>>> print(model.log_probability(['A', 'B']))
# fixed
>>> print(model.log_probability([['A', 'B']]))
```
**Note:** There is a much more detailed docstring for the `predict_proba` method, where the "dict or array-like" `X` parameter is stated to support either a "list of dicts, or...a list of lists". Perhaps including a portion of this message within the Exception string for `check_input` may be a bit more intuitive for users, especially if they are following a tutorial where a singular list is passed in. More advanced users may recognize they need to cast to `numpy.ndarray` or create an array-like list, but some newer learners may be thrown-off by this.

### 3. `BayesianNetwork.impute` appears to no longer be a valid attribute or method

`BayesianNetwork.pyx` _L196_
```python
# current
>>> print(model.impute([[None, 'A']]))
```
Results in the following error:
```
AttributeError: 'pomegranate.BayesianNetwork.BayesianNetwork' object has no attribute 'impute'
```
Manually verifying this is the case
```python
>>> "impute" in dir(BayesianNetwork)
False
```

Thanks for the work you've done on this library, I have found it incredibly useful over the last few weeks. If you'd prefer me to simply submit a PR reflecting these changes myself I'd have no problem doing that.

jmschrei:
[...]

PR #794, opened 2020-07-15
[786] Updated BayesianNetwork docstring to match current API

minelminel:
Changes reflect discussion from [#786](https://github.com/jmschrei/pomegranate/issues/786)

I have run the example code to verify its operation and have updated the printed output.


jmschrei:
Thanks!

Issue #802, opened 2020-07-26
HMM optimization


lekcyjna123:
Hi, 
today i have looked into code of HMM in looking for a way to speed up fitting this models with DiscreteDistributions and i have a question. Why should be always checked input?
https://github.com/jmschrei/pomegranate/blob/d7cf6660c67e2c92b5d7131465029823206bba2c/pomegranate/hmm.pyx#L2590-L2591
I thought that sequences doesn't change between iterations of loop? 
When i had profiled running of HMM fit function _check_input take half of fitting time, so i think that here is a place to possible optimization, because _check_input iterate over each sequence and copy each symbol.

Profile result:
[1] Improvement: 253389.81343831721     Time (s): 0.9487
[2] Improvement: 213.1332476385287      Time (s): 0.9735
[3] Improvement: 76.93109403923154      Time (s): 0.935
[4] Improvement: 44.39238843484782      Time (s): 0.9858
[5] Improvement: 28.537198789825197     Time (s): 0.948
[6] Improvement: 17.950176118465606     Time (s): 1.004
[7] Improvement: 11.243089165887795     Time (s): 0.9343
[8] Improvement: 7.315842745301779      Time (s): 0.9407
Total Training Improvement: 253789.3164752493
Total Training Time (s): 8.6180
         5061275 function calls (4967615 primitive calls) in 14.847 seconds

   Ordered by: internal time

ncalls | tottime | percall | cumtime  | percall | filename:lineno(function)
-|- |-|-|-|-
12177   | 3.759   | 0.000  |  5.109 |   0.000 | hmm.pyx:65(_check_input)
2951   | 3.527   | 0.001 |   3.527   | 0.001 | {built-in method posix.stat}
3508002 |   1.345  |  0.000   | 1.345  |  0.000 | utils.pyx:459(_check_nan)
24353/12177 |   0.626 |   0.000  |  1.852  |  0.000 | parallel.py:248(__call__)
670   | 0.626   | 0.001 |   0.626  |  0.001 | {built-in method io.open_code}
24353/12177 |   0.491 |   0.000 |   2.336 |   0.000 | _parallel_backends.py:569(__init__)
12177    |0.453   | 0.000   | 0.498   | 0.000 | hmm.pyx:2646(summarize)
24345/12177   | 0.382 |   0.000  |  0.859   | 0.000 | {method 'summarize' of 'pomegranate.hmm.HiddenMarkovModel' objects}

lekcyjna123:
[...]

PR #803, opened 2020-07-28
Optimization of HMM fitting.

lekcyjna123:
In https://github.com/jmschrei/pomegranate/issues/802 i proposed adding transcoding of input sequences in hmm.fit to avoid invoking _check_input in hmm.summarize. This fix speed up fitting HMM models about two times.

With fix:
```
[1] Improvement: 253389.81343831721	Time (s): 0.2847
[2] Improvement: 213.1332476385287	Time (s): 0.2908
[3] Improvement: 76.93109403923154	Time (s): 0.2974
[4] Improvement: 44.39238843484782	Time (s): 0.283
[5] Improvement: 28.537198789825197	Time (s): 0.2837
[6] Improvement: 17.950176118465606	Time (s): 0.2804
[7] Improvement: 11.243089165887795	Time (s): 0.2813
[8] Improvement: 7.315842745301779	Time (s): 0.2797
Total Training Improvement: 253789.3164752493
Total Training Time (s): 2.9646
```

With pomegranate 0.13.4:
```
[1] Improvement: 253389.81343831721	Time (s): 0.71
[2] Improvement: 213.1332476385287	Time (s): 0.722
[3] Improvement: 76.93109403923154	Time (s): 0.7147
[4] Improvement: 44.39238843484782	Time (s): 0.7168
[5] Improvement: 28.537198789825197	Time (s): 0.7143
[6] Improvement: 17.950176118465606	Time (s): 0.7111
[7] Improvement: 11.243089165887795	Time (s): 0.7141
[8] Improvement: 7.315842745301779	Time (s): 0.7134
Total Training Improvement: 253789.3164752493
Total Training Time (s): 6.4269
```

### DIFF
--- a/benchmarks/benchmark_hmm.py
+++ b/benchmarks/benchmark_hmm.py
@@ -79,8 +79,8 @@ def benchmark_viterbi( model, sample ):
 
 def benchmark_training( model, samples, n_jobs ):
 	tic = time.time()
-	improvement = model.train( samples, max_iterations=10, verbose=False, n_jobs=n_jobs )
-	print("{:16}: time: {:5.5}, improvement: {:5.5} ({} jobs)".format( "BW TRAINING", time.time() - tic, improvement, n_jobs ))
+	improvement = model.fit( samples, max_iterations=10, verbose=False, n_jobs=n_jobs , return_history=True)
+	print("{:16}: time: {:5.5}, improvement: {:5.5} ({} jobs)".format( "BW TRAINING", time.time() - tic, improvement[1].total_improvement[0], n_jobs ))
 
 def main():
 	n = 15
--- a/pomegranate/hmm.pyx
+++ b/pomegranate/hmm.pyx
@@ -2537,7 +2537,8 @@ cdef class HiddenMarkovModel(GraphModel):
         cdef bint check_input = alg == 'viterbi'
         cdef list X = []
         cdef list checked_sequences=[]
-        cdef list Z = []
+        cdef list W=[]
+        cdef list L=[]
 
         training_start_time = time.time()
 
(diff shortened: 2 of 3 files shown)


lekcyjna123:
I see, that not all test passed. I know already where is bug. I turned off checking input for al types of input in hmm.fit, but i don't make anything when i get DataGenerator as input data. So i hope that i will soon fix this merge request.

lekcyjna123:
I fixed my request to pass all test. I would apologize for commits which didn't passed test. I have problem with testing this files locally. Nose return me that there are errors in release 0.13.4 and this error are shown in all my version, while test with errors in GitHub passed locally.

Now i have seen that accidentally i added to this request commit which fixed benchmark for hmm and i don't how i should split it from this request to new.

jmschrei:
Nice. That's a good find. However, I don't feel comfortable simply turning that option off. Maybe we should simply make `check_input` an argument the user can pass in and have appropriate documentation telling them to turn it off for speed?

lekcyjna123:
Ok, i can try do this. 
>That way the user can turn it off when they know they have clean inputs.
We have to run `check_input` at least one time, because in this function we transcode arbitrary symbols given by user to `float64`.
>I was trying to support arbitrary inputs, including each batch being a new set of data that the mode hadn't seen before
Yes, this need running `check_input` each time, but i don't see where there is an option to pass into `fit` function data, which will be different in each step of fitting.

lekcyjna123:
I implemented new argument `multiple_check_input` for `hmm.fit()` and `hmm.from_sample()` it has default value `True`. In default configuration this functions should be the same as in pomegranate 0.13.4. When we set this argument to `False` we `check_input` only once on the beggining of `hmm.fit()`. I added also 3 new test to test_hmm.py to test situation in which `multiple_check_input=False`. This test are simple copy of basic tests of fitting models extended with this new argument.

jmschrei:
Awesome. This is a great PR. Thanks!

PR #812, opened 2020-09-08
check hmm algorithm validity

sdtaylor:
Without this check predict() will always return the `viterbi` method if the algorithm is anything but `map`


jmschrei:
Thanks!

PR #818, opened 2020-09-29
Store X_int in 32-bit precision when possible

alexhenrie:
X_int can only contain small positive integers or NaN, so double precision is unnecessary.

This saved me about 80 MB of memory on a large test net, which translates to a savings of 7 GB when we're running 88 large nets at a time. That will definitely help us stay within our memory constraints.

Test program (start with `/usr/bin/time -v`):

```
import itertools as it
import numpy as np

X = np.array(list(it.product([0,1], repeat=20)))
BayesianNetwork().from_samples(X, algorithm='chow-liu')
```

Before:

```
Elapsed (wall clock) time (h:mm:ss or m:ss): 1:24.00
Maximum resident set size (kbytes): 1841372
```

After:

```
Elapsed (wall clock) time (h:mm:ss or m:ss): 1:21.29
Maximum resident set size (kbytes): 1759564
```

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -1113,7 +1113,7 @@ cdef class BayesianNetwork(GraphModel):
 			n, d = X.shape
 
 
-		X_int = numpy.empty((n, d), dtype='float64')
+		X_int = numpy.empty((n, d), dtype='float32')
 		for i in range(n):
 			for j in range(d):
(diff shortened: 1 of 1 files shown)


jmschrei:
I guess the poor people that are trying to have 3 billion keys per column will have to use a different package.

alexhenrie:
Haha well, I'm pretty sure that you'd hit time and memory limits long before reaching 3 billion states per node :-P

PR #819, opened 2020-10-05
Use numpy.asarray instead of numpy.array where possible

alexhenrie:
numpy.array always makes a copy whereas numpy.asarray only makes a copy if necessary.

Compare:

```
$ /usr/bin/time -v python -c 'import numpy as np; np.array(np.ones(1000000))' 2>&1 | grep 'Maximum resident set'
        Maximum resident set size (kbytes): 41096
```

```
$ /usr/bin/time -v python -c 'import numpy as np; np.asarray(np.ones(1000000))' 2>&1 | grep 'Maximum resident set'
        Maximum resident set size (kbytes): 33168
```


jmschrei:
Thanks!

alexhenrie:
Thank you!

PR #822, opened 2020-10-11
Remove redundant yaml import

alexhenrie:
A straightforward simplification.


alexhenrie:
Sorry, I forgot that I already did this in #793.

PR #823, opened 2020-10-11
Eliminate unnecessary variable X_int_ptr from BayesianNetwork.summarize

alexhenrie:
Continuation of pull request #792


jmschrei:
I just pushed the latest version, so unfortunately this will have to wait until next time, but thanks!

alexhenrie:
OK. Thanks for your continued work on Pomegranate!

PR #826, opened 2020-10-13
Defining a Cython distribution for custom distance metrics

aarmey:
@mcreixell and I needed a way to define custom distances between observations when fitting a GMM. (We're clustering data where each observation is a protein peptide, and want to be able to use the PAM250 or other distance metrics to influence the clustering.) To do so, we made a "custom distribution" which can then be inherited by a Python distance class. One then makes a new column in the data that includes the indices for each observation.

We were wondering if this might be a sufficiently general to include within pomegranate. We can write a test and description if so, but wanted to check first.

### DIFF
--- a/pomegranate/distributions/CustomDistribution.pxd
+++ b/pomegranate/distributions/CustomDistribution.pxd
@@ -1,0 +1,13 @@
+# CustomDistribution.pxd
+# Contact: Aaron Meyer
+
+import numpy
+cimport numpy
+
+from .distributions cimport Distribution
+
+cdef class CustomDistribution(Distribution):
+	cdef public numpy.ndarray logWeights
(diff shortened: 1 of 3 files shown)


jmschrei:
Sorry for the late reply. I completely missed the email about this. I'm not sure what the benefit of this is versus defining your own custom distributions like in https://github.com/jmschrei/pomegranate/blob/master/tutorials/C_Feature_Tutorial_5_Custom_Distributions.ipynb, can you elaborate?

aarmey:
The difference is `_log_probability` is called for each data point, and we found that the Python overhead for this was significant. This is a minor improvement that comes with additional complexity, though. The Python version is a better general solution.

PR #829, opened 2020-10-21
Remove redundant initializations from constructors

alexhenrie:
All C variables are guaranteed to be initialized to 0 before `__cinit__` is called, and all Python variables are guaranteed to be initialized to None before `__init__` is called.

See https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#initialisation-methods-cinit-and-init

I didn't change any behavior, but I did think it was strange that the JointProbabilityTable and ConditionalProbabilityTable constructors does not use their `frozen` parameters. Is that intentional?

### DIFF
--- a/pomegranate/utils.pyx
+++ b/pomegranate/utils.pyx
@@ -63,7 +63,6 @@ cdef class PriorityQueue(object):
 	cdef dict entries
 
 	def __init__(self):
-		self.n = 0
 		self.pq = []
 		self.entries = {}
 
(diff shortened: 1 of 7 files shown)


pomegranate/utils.pyx
```diff
@@ -63,7 +63,6 @@ cdef class PriorityQueue(object):
 	cdef dict entries
 
 	def __init__(self):
-		self.n = 0
```
jmschrei, line 66: Sorry, where does n get initialized then?

Issue #845, opened 2020-11-29
[BUG] Installation via pip while missing cython fails


t4y:
On Ubuntu 20.04, using `python3 -m pip install pomegranate` to install pomegranate 0.13.5 fails similarly to what is described in #782 : building wheels fails with `x86_64-linux-gnu-gcc: error: pomegranate/distributions/NeuralNetworkWrapper.c: No such file or directory`

After Installing cpython first, the installation runs fine.

jmschrei:
Thanks for reporting this. I'm trying to revamp the wheel system to produce a more comprehensive set of wheels, so people don't encounter this issue. Sorry for the inconvenience.

jmschrei:
btw I've added this to the README for future people that encounter this issue.

jmschrei:
This should be fixed now.

PR #860, opened 2021-01-06
Tell Pip to install Cython, NumPy, and SciPy before running setup.py

alexhenrie:
[[845](https://github.com/jmschrei/pomegranate/issues/845)] Installation via pip while missing cython fails

See https://setuptools.readthedocs.io/en/latest/build_meta.html and https://github.com/pypa/setuptools/issues/1317#issuecomment-387702371


jmschrei:
Thanks for looking into this. Let's see if it works!

PR #872, opened 2021-02-09
Import pygraphviz, matplotlib, and cupy only when necessary

alexhenrie:
[According to Python's documentation](https://docs.python.org/3/reference/import.html#importsystem), import statements only import the named module once even if they are in a function that is called over and over. So moving the pygraphviz and matplotlib imports into the functions that use them should not cause any negative side effects, and it saves about 40 megabytes of RAM when they are not being used. 40 megabytes may not seem like much, but when you have 32 parallel processes all training Bayes nets, it adds up fast ;-)

Test program:

```

X = load_digits(return_X_y=True)[0][:100,:5]
BayesianNetwork().from_samples(X, algorithm='greedy')
```

Before:

```
$ /usr/bin/time -v python test.py 2>&1 | grep 'Maximum resident set'
        Maximum resident set size (kbytes): 156624
```

After:

```
$ /usr/bin/time -v python test.py 2>&1 | grep 'Maximum resident set'
	Maximum resident set size (kbytes): 116044
```

### DIFF
--- a/pomegranate/BayesianNetwork.pyx
+++ b/pomegranate/BayesianNetwork.pyx
@@ -56,14 +56,6 @@ import random
 
 from collections import defaultdict
 
-try:
-	import tempfile
-	import pygraphviz
-	import matplotlib.pyplot as plt
-	import matplotlib.image
-except ImportError:
-	pygraphviz = None
-
 DEF INF = float("inf")
 DEF NEGINF = float("-inf")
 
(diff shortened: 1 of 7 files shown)


alexhenrie:
Sorry for the mess of a pull request. It took me a while to figure out how to lazy-load cupy without breaking anything. But I've got it figured out now and that saves another 50 MB of RAM :-)

PR #873, opened 2021-02-09
Remove redundant definition of nan from BayesianNetwork.pyx

alexhenrie:
The extra line was added by commit 8ffca80b304d24f1333fd95e75f96935e6ff4993.


alexhenrie:
It looks like there is a bunch more dead code than this. I will try to make a new pull request that addresses all of it at the same time.

PR #888, opened 2021-03-19
Define other_variables as a list rather than a dict

alexhenrie:
Using a list here makes it a little more clear that all of the indices are contiguous.


jmschrei:
Won't this be wrong for all entries after `j == i`?

alexhenrie:
Ah shoot, I missed that detail. Good catch.

alexhenrie:
No wait, the if clause just makes the tuple at that index smaller. The entries in the list are definitely numbered continuously from 0 through d-1.

alexhenrie:
This sort of confusion is exactly why it should be a list :P

jmschrei:
Got it. I'm not going to merge this, though, because I'm not fond of handling small PRs like this. I don't want to encourage people to submit aesthetic PRs that don't reduce the line count or change the functionality.

PR #889, opened 2021-03-19
Skip marginal count calculation when unnecessary in discrete_score_node

alexhenrie:
Not really a big deal, but since Cython inserts division-by-zero checks here, skipping this step could shave some training time off of particularly large nets.


jmschrei:

PR #901, opened 2021-03-28
New Feature: TrueBetaDistribution

rayleizhu:
I've added TrueBetaDistribution support in this version. It will be useful to model some skewed distributions and their mixtures.


jmschrei:
I'm reluctant to merge a new distribution without adding to the suite of unit tests. Would you mind adding in unit tests similar to those of other distributions?

rayleizhu:
> I'm reluctant to merge a new distribution without adding to the suite of unit tests. Would you mind adding in unit tests similar to those of other distributions?

Do you mean adding test function in tests/test_distributions.py ?

I actually have added the function test_distributions_**true_beta**_random_sample() at Ln 717 in that file. This function just tests the sample() method, just as what is done in originally provided test_distributions_**beta**_random_sample().  

Maybe such a test is too simple, but I have no idea how to test the functionality thoroughly. Then I also write a jupyter notebook in examples/test_true_beta.ipynb to visualize the predefined and fitted distribution, it looks good. I know this is not the standard way to test the distribution class and it may be not appropriate to put it in examples. Feel free to delete it if you'd like to merge it afterward.

hhc1997:
@rayleizhu  Hello, I try to use  mixture model with TrueBetaDistribution but get parameters with Nan. Like this:
```
{
    "class" : "GeneralMixtureModel",
    "distributions" : [
        {
            "class" : "Distribution",
            "name" : "TrueBetaDistribution",
            "parameters" : [
                NaN,
                NaN
            ],
            "frozen" : false
        },
        {
            "class" : "Distribution",
            "name" : "TrueBetaDistribution",
            "parameters" : [
                NaN,
                NaN
            ],
            "frozen" : false
        }
    ],
    "weights" : [
        0.0,
        0.0
    ]
}
```
This problem may be caused by the expectation-maximization approach.  Specifically, there may have some singularities at 0 or 1 leading to this error.  For more details, you can follow the answer of [Sven Rahmann](https://www.quora.com/What-are-good-Python-packages-for-a-mixture-of-beta-distributions-fitting)
.

jmschrei:

PR #927, opened 2021-07-12
Kmeans - only do one init if init method is first-k

mparker2:
there doesn't seem to be much point in doing multiple different initialisations if using the same first-k starting points each time.


jmschrei:
Seems reasonable.

Issue #931, opened 2021-08-03
Add support to release aarch64 wheels


odidev:
Problem
---------
On aarch64, ‘pip install pomegranate’ builds the wheels from source code and then installs it. It requires the user to have a development environment installed on his system. Also, it takes some time to build the wheels than downloading and extracting the wheels from pypi. 

Resolution
-----------
On aarch64, ‘pip install pomegranate’ should download the wheels from pypi 

@jmschrei, @Team Please let me know your interest in releasing aarch64 wheels. I can help in this.

jmschrei:
Thank you for opening an issue. pomegranate has recently been rewritten from the ground up to use PyTorch instead of Cython (v1.0.0), and so all issues are being closed as they are likely out of date. Please re-open or start a new issue if a related issue is still present in the new codebase.

PR #932, opened 2021-08-03
Add linux aarch64 wheel build support

odidev:
Add linux aarch64 wheel build support. 
Related to https://github.com/jmschrei/pomegranate/issues/931 @jmschrei Could you please review this PR?

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -6,11 +6,15 @@ on: workflow_dispatch
 
 jobs:
   build_wheels:
-    name: Build wheels for ${{ matrix.os }}
+    name: Build wheels on ${{matrix.arch}} for ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


jmschrei:
Isn't this only going to produce builds on aarch64?

jmschrei:
I'm not sure that this workflow is working. There only seems to be one set of jobs spawned, not two (one for aarch64 and one for x86). Can you clarify?

jmschrei:
Oh, sorry, this is editing the other workflow. Is it possible to modify the "build" workflow first so I can see it in action?

odidev:
> Is it possible to modify the "build" workflow first so I can see it in action?

I have tried to modify the build workflow for aarch64. It looks that the flake8 command is failing for aarch64 in my forked repository but locally it is passing for aarch64 for your repository. Could you please suggest some pointers? 

**GHA Logs:** https://github.com/odidev/pomegranate/runs/4088748735?check_suite_focus=true

odidev:
@jmschrei Could you please review this PR? Thanks in advance.

jmschrei:

PR #946, opened 2021-11-01
[WIP] Fix wheel issues

jmschrei:
There appear to be issues related to building the scipy wheel for py3.10 and also for deployment.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -30,6 +30,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install flake8 pytest
+        pip install openblas
         pip install -r dev-requirements.txt
         python setup.py install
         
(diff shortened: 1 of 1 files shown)


shyamd:
Quick look at this PR and it appears is failing not on the build of SciPy now, but on activating the Conda Environment. It looks like initially you were trying to build scipy on Mac and Linux. Can you just install SciPy via Conda or pip rather than trying to install it?

jmschrei:
Yes, initially it was just using pip but couldn't find blas. I couldn't pip or conda install openblas because of permission errors on the mac build so I went down that rabbit hole, couldn't get anything to work, and now I'm just not supporting Py3.10.

shyamd:
Uggh, yeah. Those components can be fussy on different systems. SciPy has a pretty notorious build system that they recently switched over to meson: https://labs.quansight.org/blog/2021/07/moving-scipy-to-meson/

It might be something to consider if you need complicated build language. Also, distutils is going away by 3.12 :P Thanks python devs. >:| 

I can take a look at it tomorrow morning, since its night for me right now. Cool project btw!

Issue #947, opened 2021-11-08
[BUG] GeneralMixtureModel of ICD of DiscreteDistributions and numeric distributions does not fit


lmcinnes:
First of all, thanks for all the work on pomegranate!

Possibly related to #402 there seems to be an issue (or a failure of understanding on my part -- also likely) with a ``GeneralMixtureModel`` of ``IndependentComponentDistribution``s which ``DiscreteDistribution`` with another continuous/numeric distribution (``PoissonDistribution``, ``GammaDistribution`` etc.). Initialization of the model works fine, but upon trying to ``fit`` there are complaints about ``KeyError``s. Specifically I get the error:

```python-traceback
Traceback (most recent call last):
  File "pomegranate/utils.pyx", line 429, in pomegranate.utils._check_input
ValueError: could not convert string to float: 'A'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pomegranate/gmm.pyx", line 220, in pomegranate.gmm.GeneralMixtureModel.fit
  File "pomegranate/gmm.pyx", line 231, in pomegranate.gmm.GeneralMixtureModel.fit
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/parallel.py", line 1043, in __call__
    if self.dispatch_one_batch(iterator):
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/parallel.py", line 861, in dispatch_one_batch
    self._dispatch(tasks)
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/parallel.py", line 779, in _dispatch
    job = self._backend.apply_async(batch, callback=cb)
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/_parallel_backends.py", line 208, in apply_async
    result = ImmediateResult(func)
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/_parallel_backends.py", line 572, in __init__
    self.results = batch()
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/parallel.py", line 262, in __call__
    return [func(*args, **kwargs)
  File "/home/leland/anaconda3/envs/pomegranate/lib/python3.9/site-packages/joblib/parallel.py", line 262, in <listcomp>
[...]

PR #948, opened 2021-11-09
Fix for #947: Be more flexible in ICD keymaps

lmcinnes:
If an ICD has a mixture of ``DiscreteDistribution`` and other distribution types then the discrete distributions have a keymap for encoding, and the other distributions get an empty keymap (based on initialization of keymaps per component as empty dicts). When data is then passed in the check_array uses keymaps to encode the data, but fails for the non-``DiscreteDistribution`` columns since they have empty keymaps. If the column has data for a non-``DiscreteDistribution`` it is numeric and doesn't require encoding -- the entire column can simply be cast to ``float64``. This PR makes a minimal alteration to work by columns first, and, in the case of an empty keymap attempts to cast the entire column to ``float64``. This resolves the issues seen in #947 . A test of this is also added.

Note: running the test suite locally I have a failure for ``tests.test_naive_bayes.test_discrete_distribution``, however this test also fails on the current master, and does not seem related to these changes. All other tests passed.

### DIFF
--- a/pomegranate/utils.pyx
+++ b/pomegranate/utils.pyx
@@ -440,9 +440,15 @@ def _check_input(X, keymap=None):
 					X_ndarray[i] = keymap[0][X[i]]
 				X_ndarray = X_ndarray.reshape(-1, 1)
 			else:
-				for i in range(X.shape[0]):
-					for j in range(X.shape[1]):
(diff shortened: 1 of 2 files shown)


jmschrei:
Thanks for the contribution and sorry for the delay -- busy this week. I'll look this over soon and get it merged!

PR #950, opened 2021-12-01
Temporary fix for GH Actions and Py3.10

shyamd:
Hopefully this works on the main repo and not just my fork.
I think the big issue is actually `nose` which is no longer supported in Python 3.10

The other thing to note is that `cibuildwheel` can actually integrate in testing, so it might make more sense to use that to both test and build for release so that you are less likely to encounter some problem that was not checked in the CI.


jmschrei:

Issue #937, opened 2021-08-21
Multivariate HMM issues while training


mellorf:
Dear Jmschrei,

Tks a lot for such an useful library!

I found an issue while training my HMM as follows:

```
import numpy as np
from pomegranate import *
data=np.array([[[20.0,30.0,10.0],[21.0,29.0,11.0],[19.0,31.0,9.5],
                [1.0,1.0,1.0],[1.1,1.1,1.1],[110.0,600.0,100.0],
                [105.0,595.0,99.9],[20.0,30.5,10.1]], 
              [[21.0,30.0,10.0],[21.0,29.0,11.0],[19.0,31.0,9.5],
                [2.0,1.0,1.0],[1.1,1.1,1.1],[110.0,600.0,100.0],
                [107.0,595.0,99.9],[20.0,30.5,10.1]]])

GMMHMM=HiddenMarkovModel.from_samples(MultivariateGaussianDistribution, n_components=3,
                                      X=data, algorithm='baum-welch', inertia=0.01, 
                                      edge_inertia=True, distribution_inertia=0.01, 
                                      min_iterations=2, max_iterations=1e8, n_init=10, 
                                      lr_decay=1e-3, state_names=['Estado1', 'Estado2', 'Estado3'],
                                      verbose= True, n_jobs=1)
```

The code above works beautifully, however when I use the following code to train it:

```
Y = [[0],[1]]
fitted = GMMHMM.fit(data, labels=Y, algorithm='labeled')
```

it produces the following error:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
KeyError: 0

Exception ignored in: 'pomegranate.hmm.HiddenMarkovModel._labeled_summarize'
KeyError: 0
```

I am sure the error is in Y, so I tried another possibility which has also failed:

```
Y = [[0,0,1,0,0,0,0,0],[1,1,0,1,1,1,1,1]]
fitted = GMMHMM.fit(data, labels=Y, algorithm='labeled')
```

I kindly thank you once more for your support!

Best,

Rodrigo

mellorf:
Solved! Tks!!

```
import numpy as np
from pomegranate import *
data=np.array([[[20.0,30.0],[3.0,7.0],[3.5,7.1],[20.5,30.2],[200.0,400.0],[201.0,401.0],[3.1,6.9],[20.1,30.1]]])
y=[0,1,1,0,2,2,1,0]
[...]

Issue #549, opened 2019-02-11
Cannot capture "ignored Exception"


asapegin:
My sample code always returns an exception even though I try to catch it:

<pre>
from pomegranate import *
a = ["A","B","C"]
b=["A","B","B"]
c=["A","B","C"]
d=["A","C","C"]
tdata = [a,b,c,d]
thmmodel = HiddenMarkovModel()
thmm = thmmodel.from_samples(DiscreteDistribution, 5, tdata, max_iterations=10, n_init=1)
test2 = ["A","B","D"]
try:
  lprob = thmm.log_probability(test2)
  print("lprob=",lprob)
except Exception as e:  
  print("Exception catched: ",e)
</pre>

I understand that there should be an exception if a symbol was not defined in the distribution. In my program, I want to handle such cases and therefore try to catch an exception. However, it is still not catched, rather it outputs the following:
<pre>
Exception ignored in: 'pomegranate.hmm.HiddenMarkovModel.log_probability'
Traceback (most recent call last):
  File "pomegranate/hmm.pyx", line 100, in pomegranate.hmm._check_input
ValueError: Symbol 'D' is not defined in a distribution
lprob= 0.0
</pre>

After this, a Value Error is raised although I tried to catch it:

<pre>
ValueError: Symbol 'D' is not defined in a distribution
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/databricks/python/lib/python3.5/site-packages/pomegranate/hmm.cpython-35m-x86_64-linux-gnu.so in pomegranate.hmm._check_input()

ValueError: Symbol 'D' is not defined in a distribution
</pre>

Do you have any ideas how to catch it? I run the code in Databricks Runtiome 4.3, Spark 2.3.1, Python 3

jmschrei:
I'm not aware of how to do that. Unfortunately I think that using Cython kind of messes up error handling. Sorry about that.

PR #952, opened 2021-12-17
Better exception handling

Dobatymo:
Added exception handling to `HiddenMarkovModel.log_probability` so Python exceptions can be caught and are not just logged and ignored.

This problem is mentioned here https://github.com/jmschrei/pomegranate/issues/549 or https://github.com/jmschrei/pomegranate/issues/937

I am not sure how other cpdef functions which return Python objects and not native types need to be handled.


jmschrei:

PR #1006, opened 2022-11-15
Switch to pytest from unsupported nose

mr-c:
We need this for Debian as nose isn't working with Python 3.11

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1024045

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -47,4 +47,4 @@ jobs:
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
         
     - name: Test with pytest
-      run: python setup.py test
+      run: python -m pytest
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ install:
 	python setup.py install
 
 test:
-	python setup.py test
+	python3 -m pytest
 
 bigclean: nbclean
 	rm -rf build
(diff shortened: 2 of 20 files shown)


jmschrei:
Oh, great, thanks! Unfortunately, I'm in the middle of rewriting pomegranate using torch and the unit tests for that are being written in pytest. I don't have the time to continue to develop pomegranate separate from that rewrite. I am anticipating that the first public release will be next week. 

btw I had been using the nose assert functions with pytest but here you remove them. I haven't actually checked with py3.11 yet -- is the nose module in its entirety broken?

mr-c:
> Oh, great, thanks! Unfortunately, I'm in the middle of rewriting pomegranate using torch and the unit tests for that are being written in pytest. I don't have the time to continue to develop pomegranate separate from that rewrite. I am anticipating that the first public release will be next week.

Okay. These patches work for us, so maybe another distribution will find them useful.

> btw I had been using the nose assert functions with pytest but here you remove them. I haven't actually checked with py3.11 yet -- is the nose module in its entirety broken?

Yeah, `nose` is relying on `getargspec` being in the `inspect` module, but that is no longer there for Python 3.11 (it was deprecated in Python 3.0)
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/usr/lib/python3/dist-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/usr/lib/python3/dist-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/usr/lib/python3/dist-packages/nose/util.py", line 453, in try_run
    inspect.getargspec(func)
    ^^^^^^^^^^^^^^^^^^
AttributeError: module 'inspect' has no attribute 'getargspec'
```

Given that `nose` has had no [releases since 2015](https://pypi.org/project/nose/#history) and [no commits since 2016](https://github.com/nose-devs/nose/commits/master) I don't expect to ever have a `nose` release with Python 3.11 compatibility

jmschrei:
Got it. Thanks for the info and for taking the time to help me out by rewriting these tests.

jmschrei:

PR #1007, opened 2022-11-25
Missing variable assignment

saulpierotti-ebi:
`name` was not assigned to `self.name` and as a consequence it was not possible to create checkpoint files with name different from model.name


jmschrei:
Howdy. Thanks for pointing this issue out. However, I'm not currently accepting changes to pomegranate because I am about to release a ground-up rewrite of the package in torch. I'm currently staging it at https://github.com/jmschrei/torchegranate.

jmschrei:

PR #1019, opened 2023-04-17
Fix hmm fit shape check

### LINKED ISSUE #1018: [BUG] Check for prior shape in HMM.fit is wrong
**Describe the bug**
For the HMM.fit method priors (if I understood correctly), should have the shape `(n, length, n_states)`, if the data has the shape `(n, length, dim)`.
However, in `partition_sequences` (https://github.com/jmschrei/pomegranate/blob/c77f967a2b66505b42a4fc4063fcf1d26406a9a5/pomegranate/_utils.py#L410), the code checks that prior has the exact same shape as `X`.
I think this is incorrect.
The same bug will occure for other input shapes to fit as well.

**To Reproduce**
To reproduce, we can simple change the last line of the semi-supervised example to call `fit` instead of `predict_proba`:

```python
from pomegranate.hmm import DenseHMM
from pomegrante.distributions import Normal

d1 = Normal([1.0], [1.0], covariance_type='diag')
d2 = Normal([3.0], [1.0], covariance_type='diag')

model = DenseHMM([d1, d2], [[0.7, 0.3], [0.3, 0.7]], starts=[0.4, 0.6])

priors = torch.ones(1, 10, 2) / 2
priors[0, 5, 0], priors[0, 5, 1] = 0, 1
priors

# The following line is changed compared to the example
model.fit(X, priors=priors)
# -> Raises Value Error
```

AKuederle:
This is a proposal on how to fix #1018 .

It basically forwards the number of states to `partition_sequence` to check the correct shape for the priors.

So far I only tested that this will not raise an error. I did not check, if the priors are used correctly in training.

Note: This also includes the fix proposed in #1017. If this PR is merged #1017 can be closed. If #1017 is merged separably, this PR needs to be rebased

### DIFF
--- a/pomegranate/_utils.py
+++ b/pomegranate/_utils.py
@@ -355,7 +355,7 @@ def _initialize_centroids(X, k, algorithm='first-k', random_state=None):
 		return selector.fit_transform(X)
 
 
-def partition_sequences(X, sample_weight=None, priors=None):
+def partition_sequences(X, sample_weight=None, priors=None, priors_last_dim=None):
 	"""Partition a set of sequences into blobs of equal length.
 
 	This function will take in a list of sequences, where each sequence is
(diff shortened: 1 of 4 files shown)


jmschrei:
Nice! Very thorough PR. Happy to merge this, but would you mind changing the variable name from `priors_last_dim` to `n_dists`? That would be semantically more helpful.

AKuederle:
> Nice! Very thorough PR. Happy to merge this, but would you mind changing the variable name from `priors_last_dim` to `n_dists`? That would be semantically more helpful.

Done :)

jmschrei:
Thanks!

PR #1027, opened 2023-04-20
Added `Distribution` to public exports

AKuederle:


jmschrei:
I'm going to close this for now. Thanks for looking closely at the code and making this suggestion, though. I'll keep it in mind and if it makes sense in the future I'll add it in, though!

Issue #1036, opened 2023-05-25
[BUG] Normal(covariance_type='sphere') crashes when fitting


gerwang:
**Describe the bug**
Currently Normal distributions with `covariance_type='sphere'` are not properly considered in fitting. When  `covariance_type='sphere'`, its `covs` becomes a scalar torch tensor, which needs special care.

**To Reproduce**

Running the following code

```python
from pomegranate.distributions import Normal
import torch

d = Normal(covariance_type='sphere').fit(torch.randn((10, 3)))
```

raises the following error:

```
File .../pomegranate/pomegranate/distributions/normal.py:169, in Normal._reset_cache(self)
    166 	self.register_buffer("_log_sigma_sqrt_2pi", _log_sigma_sqrt_2pi)
    167 	self.register_buffer("_inv_two_sigma", _inv_two_sigma)
--> 169 if any(self.covs < 0):
    170 	raise ValueError("Variances must be positive.")
...
TypeError: iteration over a 0-d tensor
```

PR #1037, opened 2023-05-25
fit normal distribution when covariance_type == 'sphere'

gerwang:
This is a proposal on how to fix https://github.com/jmschrei/pomegranate/issues/1036.

It makes some operations more generic to consider both vector tensor and scalar tensor.

It also implements estimating spherical covariances based on https://github.com/scikit-learn/scikit-learn/blob/364c77e047ca08a95862becf40a04fe9d4cd2c98/sklearn/mixture/_gaussian_mixture.py#L258.


jmschrei:
Thanks! Sorry for the delay.

PR #1071, opened 2023-11-12
Efficient Tensor conversion from list of numpy arrays

josalhor:
While playing around with pomegranate v1 I run into this warning:

`
/home/josalhor/Desktop/data-synt/venv/lib/python3.11/site-packages/pomegranate/_utils.py:62: UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor. (Triggered internally at ../torch/csrc/utils/tensor_new.cpp:261.)
`

I think my pull requests fixes that warning and should speed up the casting.

### DIFF
--- a/pomegranate/_utils.py
+++ b/pomegranate/_utils.py
@@ -56,7 +56,10 @@ def _cast_as_tensor(value, dtype=None):
 			return value
 		else:
 			return value.type(dtype)
-
+			
+	if isinstance(value, list) and all(isinstance(v, numpy.ndarray) for v in value):
+		value = numpy.array(value)
+		
 	if isinstance(value, (float, int, list, tuple, numpy.ndarray)):
 		if dtype is None:
 			return torch.tensor(value)


jmschrei:
Would you mind undoing the version bump? Usually I'll only do a bump after a few changes, and have a corresponding release when I do it.

josalhor:
Absolutely no problem. Actually, I did the version bump because GitHub Actions reported a cancellation but no failure and wanted to retest. Will undo the commit.

jmschrei:
Thanks!

pomegranate/_utils.py
```diff
@@ -56,7 +56,10 @@ def _cast_as_tensor(value, dtype=None):
 			return value
 		else:
 			return value.type(dtype)
-
+			
+	if isinstance(value, list) and all(isinstance(v, numpy.ndarray) for v in value):
+		value = numpy.array(value)
```
jkleckner, line 61: This doesn't seem to have a return statement in all cases?

jmschrei: It's saying that if `value` is a list of numpy arrays, cast `value` as a single numpy array. L63 will then take that new numpy array and operate on it, but more efficiently than if it were a list of arrays.

jkleckner: Doh!

josalhor:
Did this PR break the build?

jmschrei:
No. There are some unit tests that stochastically raise NaNs. When I manually re-run them they run successfully. 100% of the time these same tests run successfully on my local machine so I'm not sure what their source is.

PR #1076, opened 2024-01-19
Fix typos

FBruzzesi:
Not super exciting PR, but I found some typos in the codebase fast and easy to fix 😊

Edit: Test are failing with some precision difference in the order of 1e-6, which is the default of `numpy.testing.assert_array_almost_equal`


pomegranate/_utils.py
```diff
@@ -78,7 +78,7 @@ def _cast_as_parameter(value, dtype=None, requires_grad=False):
 
 
 def _update_parameter(value, new_value, inertia=0.0, frozen=None):
-	"""Update a parameters unles.
+	"""Update a parameter unles.
```
FBruzzesi, line 81: I am not sure what the docstring should be here

jmschrei: It seems perfectly clear. You update a parameter unles.

jmschrei:
Reading this PR is embarrassing.

FBruzzesi:
> Reading this PR is embarrassing.

It happens to everyone, no need to feel embarrassed! If I may suggest a like-able tool to semi automate this process: [typos](https://github.com/crate-ci/typos) is (very) fast, flexible and customizable.

PR #1104, opened 2024-06-15
Fixed sample_weights not casted to tensor in Normal distribution

dmg99:
In the file distributions.normal, in the summarize function the input data was casted to tensor and the sample_weights were not. This raised an error later when computing self._xw_sum


jmschrei:
Thanks!

